Issue #4, opened 2016-07-19
Notebook explaining an RNA-seq classifier


stephenshank:
There are Jupyter notebooks on [acquiring](https://github.com/dhimmel/cancer-data/blob/master/1.TCGA-download.ipynb) and [cleaning](https://github.com/dhimmel/cancer-data/blob/ffe66ab26000379adcd7138b8ff39920d4692ef1/2.TCGA-process.ipynb) data from TCGA. Is there one that has a outline of the type of classifier that Cognoma will run on this data for our users, and if not, would it be much trouble to create one?

Ideally, it would continue from 1) data acquisition and 2) cleansing (done above), and go on to 3) feature extraction, 4) learning, and 5) interpretation. Perhaps I have missed some steps?

A paper of @cgreene, @dhimmel or @gwaygenomics might also suffice, but it could be nice for everyone to have some code to play around with. Thoughts?

dhimmel:
I think a notebook would serve as a great example here.

You could use the [subsetted files available here](https://github.com/dhimmel/cancer-data/tree/process/data/subset). For example, if you wanted to fit a model for all samples, but only a subset of the genes, you can read the datasets with:

``` python
import pandas

base_url = 'https://github.com/dhimmel/cancer-data/raw/ffe66ab26000379adcd7138b8ff39920d4692ef1/data/subset/'

# Read expression data (features)
expression_df = pandas.read_table(base_url + 'expression-matrix-all-samples.tsv')

# Read mutation data (to create the outcome/status)
mutation_df = pandas.read_table(base_url + 'mutation-matrix-all-samples.tsv')
```

For the mutation data, it may be necessary to ask whether a sample has a mutation in any of the genes (because the data is so sparse).

Making this notebook sounds like a good task. I think we could even have multiple people make their own. Anyone want to give it a try?

cgreene:
@gwaygenomics just put up an example via PR #10 . Once that gets approved we'll have an example notebook. @stephenshank: since you posted the issue, do you want to review the PR?

stephenshank:
Would be my pleasure! I am not sure whether or not I have the privileges to assign this to myself. At first glance, it looks very nice... thanks @gwaygenomics !

dhimmel:
Closing as #10 is now merged.

[...]

PR #10, opened 2016-07-25
adding machine learning example

gwaybio:
Here is an example classifier for what Project Cognoma will eventually be able to do for any gene, any tissue, and any machine learning classifier.

The classifier is a logistic regression classifier with an elastic net penalty. This simple classifier, with little tuning, predicts 'TP53' mutation status with 97% accuracy in a hold out set. TP53 is the most abundantly mutated gene and massively affects transcription - therefore, it is a good positive control


dhimmel:
`pd.Series(Y['TP53'], index=Y.index)` is equivalent to `Y['TP53']` I believe. Also should be `Y[GENE]`?

gwaybio:
Yes, nice catch @dhimmel - I will update the PR to fix this change in the next commit

dhimmel:
Assigning @stephenshank to review this pull request. See https://github.com/cognoma/machine-learning/issues/4#issuecomment-235098910

dhimmel:
I wouldn't hardcode the gene into the variable name (`Y_p53`). Perhaps you could switch `Y_p53` to `y`. I think I've seen lowercase for 1d and uppercase for 2d somewhere... but I forget.

stephenshank:
@gwaygenomics This is very helpful and instructive. In addition to @dhimmel's comments all I would say to add right now are the imports
`from sklearn import preprocessing`
`import matplotlib.pyplot as plt`
for someone who wants to run this and does not automatically load these modules in their environments. After these additions, this notebook ran just fine on my system.

gwaybio:
@stephenshank good point, I will add those to the import and also make the changes @dhimmel suggested. Thanks for the comments!

dhimmel:
@gwaygenomics, you should "Restart & Run All" on your notebook before committing it. Keep it fresh. This way you won't get any of these errors which result from non-linear execution and changed code. This is something that we can have pull request reviewers look out for in the future.

gwaybio:
Ok @stephenshank and @dhimmel - PR is updated! I addressed the `Y_p53` variable, the `y = Y[GENE]` comment, added the two imports, and restarted and reran the kernel. Thanks!

dhimmel:
Great will merge now, so we're ready for tonight's meetup!

PR #25, opened 2016-08-03
Using pipeline in the TCGA-MLexample, add feature selection

yl565:
I modified the example to use the following pipeline:
median absolute deviation (MAD) feature selection -> z-score scaling -> Grid_search with SGDClassifier
Since pipeline is used, all the feature extraction steps (feature selection/scaling) are estimated using only x_train. The mask of selected feature and the estimated mean/std (for scaling)  is then applied to x_test.

To reduce computation time, I used the top 500 features with the largest MAD. Higher performance may be achieved with more features included (grid search the optimal number of features may also be an option).

(I noticed in the example x_train and x_test are standardized separately which works fine if x_test contains sufficient samples. From a practical point of view, I think it would works better to estimate the mean/std from x_train and apply it to standardize x_test)


dhimmel:
@yl565, awesome! I've never used an sklearn pipeline, but I think it's the right move.

Use the conda environment in #15 (even though it's not merged yet) and remove the pandas package list.

Export notebook to script so we can comment on specific lines.

Rather than create a new notebook, save to original name. Let me know if you disagree.

dhimmel:
You can use the following command to export all notebooks to scripts:

``` sh
jupyter nbconvert --to=script --FilesWriter.build_directory=scripts *.ipynb
```

yl565:
Ok, it takes me a while to setup the environment and commit all the changes. Still learning how to use git.  Is everything working? Also, can I ask what command we should use to commit all changes (including delete some files)

dhimmel:
> what command we should use to commit all changes (including delete some files)

You can do `git add --all` which adds and removes everything. This is sometimes too large, as there are changes you don't want to commit. You can therefore do `git rm deleted_file_name` and then `git add new_file_name`.

dhimmel:
> Still learning how to use git. Is everything working?

Everything is working. One issue is that you made your changes on top of ab862e0d7d00c359b8e806eb96151f2285b84c8f (an old repo version) rather than ae27311b684c31bacfee7d0dd2be9f418d7301fc (the most recent version). Therefore, you're missing these four commits:

```
ae27311b684c31bacfee7d0dd2be9f418d7301fc Use grid_search in notebook and add visualization (#18)
1781a412c9ad818c9cb290e97bfbf29a2ed6075c Ignore Jupyter notebook checkpoints (#17)
7ab14ad44860c904abc4ecc768a13508b4f79575 Export notebook to .py file for easy diff viewing (#16)
892c6994d0f57c671611781d1b96cbba31ee3c7e adding machine learning example (#10)
```

There may be a way for you to rebase and redo your changes on top of the most recent master. For now this isn't a huge deal, but it means that the pull request cannot be automatically merged. (I can still merge it manually though).

yl565:
Thanks for the tips. I synced my fork to master and updated the two files, looks like it can be auto-merged now?

dhimmel:
Yeah, looks like what you did did the trick. The diff view on `scripts/1.TCGA-MLexample.py` is now really helpful. I have a few more comments to make, which I'll do now.

scripts/1.TCGA-MLexample.py
```diff
@@ -11,33 +11,36 @@
...
 import pandas as pd
+import numpy as np
 import matplotlib.pyplot as plt
 import seaborn as sns
 from sklearn import preprocessing, grid_search
 from sklearn.linear_model import SGDClassifier
 from sklearn.cross_validation import train_test_split
 from sklearn.metrics import roc_auc_score, roc_curve
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import Imputer, StandardScaler
```
dhimmel: `Imputer` is never used. let's keep the imports minalist.

scripts/1.TCGA-MLexample.py
```diff
@@ -72,84 +75,96 @@
...
 
-# In[13]:
+# In[14]:
 
-def scale_df(df):
-    """Standardize colummns by scaling to z-scores"""
-    scaled_matrix = preprocessing.scale(df)
-    return pd.DataFrame(scaled_matrix, index=df.index, columns=df.columns)
+def fs_mad(x, y):
+    mad = lambda a: np.median(np.absolute(a - np.median(a)))
```
dhimmel: Even though MAD is simple, I prefer not to reimplement [`statsmodels.robust.scale.mad`](http://statsmodels.sourceforge.net/devel/generated/statsmodels.robust.scale.mad.html). `statsmodels=0.6.1` [is available](https://github.com/cognoma/machine-learning/pull/15/files#diff-b7cb75ca59993a879e29bed8cd8469b8R11) in our environment in #15.

Also make sure to add a docstring explaining what the function does.

Are you returning fake p-values? Can you return a vector of None or nan instead?

scripts/1.TCGA-MLexample.py
```diff
@@ -72,84 +75,96 @@
...
-X_train, X_test = scale_df(X_train), scale_df(X_test)
 
+# ## Define pipeline and Cross validation model fitting
 
-# ## Cross validation model fitting
+# In[15]:
 
-# In[14]:
+# Include loss='log' in param_grid doesn't work with pipeline somehow
+clf = SGDClassifier(random_state=0, class_weight='balanced', loss='log')
```
dhimmel: I kind of think loss should go in `param_grid`. For example, you may want to specify `loss='hinge'` and set a totally different `alpha` list? What's are your thoughts on which attributes we set here versus `param_grid`?

scripts/1.TCGA-MLexample.py
```diff
@@ -72,84 +75,96 @@
...
 
 # joblib is used to cross-validate in parallel by setting `n_jobs=-1` in GridSearchCV
 # Supress joblib warning. See https://github.com/scikit-learn/scikit-learn/issues/6370
 warnings.filterwarnings('ignore', message='Changing the shape of non-C contiguous array')
+clf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')
+pipeline = make_pipeline(
+              feature_select,  # Feature selection
```
dhimmel: Style: switch to a single indent?

scripts/1.TCGA-MLexample.py
```diff
@@ -72,84 +75,96 @@
...
+# In[16]:
 
-get_ipython().run_cell_magic('time', '', "clf = SGDClassifier(random_state=0, class_weight='balanced')\nclf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')\n# Fit the model (the computationally intensive part)\nclf_grid.fit(X=X_train, y=y_train)\nbest_clf = clf_grid.best_estimator_")
+get_ipython().run_cell_magic('time', '', '# Fit the model (the computationally intensive part)\npipeline.fit(X=X_train, y=y_train)\nbest_clf = clf_grid.best_estimator_')
 
 
-# In[16]:
+# In[17]:
+
+# Check the list of selected features
```
dhimmel: I don't find this cell helpful for the final notebook. i.e. great for development but remove once you see it worked as intended? In other words, I don't think the notebook viewer is gaining much from this cell and the output is visually distracting.

scripts/1.TCGA-MLexample.py
```diff
@@ -283,8 +300,13 @@ def get_threshold_metrics(y_true, y_pred):
 ax = sns.distplot(predict_df.query("status == 1").decision_function, hist=False, label='Positives')
 
 
-# In[30]:
+# In[32]:
 
 ax = sns.distplot(predict_df.query("status == 0").probability, hist=False, label='Negatives')
 ax = sns.distplot(predict_df.query("status == 1").probability, hist=False, label='Positives')
 
+
```
dhimmel: When you think you're done with everything, clean up for final commit:
- remove this empty cell
- restart notebook and run all so all cells are in sequence from 1 to n.
- rexport to script

dhimmel:
Those are all the comments I have.

scripts/1.TCGA-MLexample.py
yl565: Thanks for your comments. I agreed with most of them. For the param_grid though, it seems that if loss is in param_grid as in the following code

``` python
param_grid = {
    'loss': ['log'],
    'penalty': ['elasticnet'],
    'alpha': [10 ** x for x in range(-6, 1)],
    'l1_ratio': [0, 0.05, 0.1, 0.2, 0.5, 0.8, 0.9, 0.95, 1],
}
clf = SGDClassifier(random_state=0, class_weight='balanced')
clf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')
pipeline = make_pipeline(
              feature_select,  # Feature selection
              StandardScaler(),  # Feature scaling
              clf_grid)
pipeline.fit(X=X_train, y=y_train)
pipeline.predict_proba(X)
```

I will get the following error message

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-28-f384560df3ec> in <module>()
----> 1 pipeline.predict_proba(X)

/home/yichuanliu/Programs/anaconda3/envs/cognoma-cancer-data/lib/python3.5/site-packages/sklearn/utils/metaestimators.py in __get__(self, obj, type)
     33             # delegate only on instances, not the classes.
     34             # this is to allow access to the docstrings.
---> 35             self.get_attribute(obj)
     36         # lambda, but not partial, allows help() to work with update_wrapper
     37         out = lambda *args, **kwargs: self.fn(obj, *args, **kwargs)

/home/yichuanliu/Programs/anaconda3/envs/cognoma-cancer-data/lib/python3.5/site-packages/sklearn/utils/metaestimators.py in __get__(self, obj, type)
     33             # delegate only on instances, not the classes.
     34             # this is to allow access to the docstrings.
---> 35             self.get_attribute(obj)
     36         # lambda, but not partial, allows help() to work with update_wrapper
     37         out = lambda *args, **kwargs: self.fn(obj, *args, **kwargs)

/home/yichuanliu/Programs/anaconda3/envs/cognoma-cancer-data/lib/python3.5/site-packages/sklearn/linear_model/stochastic_gradient.py in predict_proba(self)
    754         http://jmlr.csail.mit.edu/papers/volume2/zhang02c/zhang02c.pdf
    755         """
--> 756         self._check_proba()
    757         return self._predict_proba
    758 

/home/yichuanliu/Programs/anaconda3/envs/cognoma-cancer-data/lib/python3.5/site-packages/sklearn/linear_model/stochastic_gradient.py in _check_proba(self)
    719         if self.loss not in ("log", "modified_huber"):
    720             raise AttributeError("probability estimates are not available for"
--> 721                                  " loss=%r" % self.loss)
    722 
    723     @property

AttributeError: probability estimates are not available for loss='hinge'
```

My guess is this is probably a bug in grid_search. I think one solution is to specify the loss  and other parameters that do not need to be grid searched outside of param_grid:

```
SGD_loss = 'log'
SGD_penalty = 'elasticnet'
n_feature_kept = 500
param_grid = {
    'alpha': [10 ** x for x in range(-6, 1)],
    'l1_ratio': [0, 0.05, 0.1, 0.2, 0.5, 0.8, 0.9, 0.95, 1],
}
feature_select = SelectKBest(fs_mad, k=n_feature_kept)
clf = SGDClassifier(random_state=0, class_weight='balanced', penalty=SGD_penalty, loss=SGD_loss)
clf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')
pipeline = make_pipeline(
              feature_select,  # Feature selection
              StandardScaler(),  # Feature scaling
              clf_grid)
```

Let me know your opinion

dhimmel: I had the same error (`AttributeError: probability estimates are not available for loss='hinge'`), which is a bug because we're using log loss not hinge. I noticed the bug went away I used `clf_grid.best_estimator_.predict_prob` rather than `clf_grid.predict_prob`. Really weird. Do you have access to `best_estimator_` through the pipeline?

dhimmel: It's odd because `clf_grid.best_estimator_.predict_prob` should just be calling `clf_grid.predict_prob` ([source](https://github.com/scikit-learn/scikit-learn/blob/dcca31c40ba8cf254c3653201fb04ade98aaa7bd/sklearn/grid_search.py#L445))

yl565: Since clf_grid is passed as a handle, after fitting the pipeline we can just call clf_grid.best_estimator_. But then we need to manually apply feature_select and scaling before calling predict_prob which lost the point of using a pipeline

dhimmel: Okay, let's proceed with your workaround.

yl565: Sounds good to me. 
It seems that the bug come from the decorator @if_delegate_has_method(delegate='estimator') from the BaseSearchCV class

``` python
    @if_delegate_has_method(delegate='estimator')
    def predict_proba(self, X):
        """Call predict_proba on the estimator with the best found parameters.

        Only available if ``refit=True`` and the underlying estimator supports
        ``predict_proba``.

        Parameters
        -----------
        X : indexable, length n_samples
            Must fulfill the input assumptions of the
            underlying estimator.

        """
        return self.best_estimator_.predict_proba(X)
```

If I remove the decorator, everything works as expected:

```
pipeline.predict_proba(X)
Out[4]: 
array([[ 0.42683613,  0.57316387],
       [ 0.28014646,  0.71985354],
       [ 0.29039856,  0.70960144],
       ..., 
       [ 0.72043756,  0.27956244],
       [ 0.9025268 ,  0.0974732 ],
       [ 0.62758795,  0.37241205]])
```

dhimmel: Wow great detective work! If you think you know how to fix the source code, I'd submit a pull request to sklearn (it's always cool to become a contributor on a repo with 1000+ stars!). Otherwise you could file an issue (or I could).

For now let's go ahead with a workaround of your choosing.

yl565:
OK, I updated the files. Also fixed a bug so that its now printing the correct number of features

```
In [13]:

# Typically, this can only be done where the number of mutations is large enough
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
'Size: {:,} features, {:,} training samples, {:,} testing samples'.format(len(X.columns), len(X_train), len(X_test))
Out[13]:
'Size: 20,501 features, 6,935 training samples, 771 testing samples'
```

dhimmel:
Great work with this pull request. I'm going to merge it.

dhimmel:
After I merged this I realized that 34225ccdefa191287ca153fc14c73bb4eaa6706d (which all of your commits were squashed into) modified the mode of several text files to executables. This is something we will want to change back. Will look more into it later

scripts/1.TCGA-MLexample.py
dhimmel: Just wanted to mention that @yl565 has gone ahead and filed the an issue (https://github.com/scikit-learn/scikit-learn/issues/7155) and pull request (https://github.com/scikit-learn/scikit-learn/pull/7159) on this bug with `sklearn`. Awesome work!

Issue #27, opened 2016-08-07
Claim an sklearn algorithm to implement and troubleshoot


dhimmel:
In the August 26 meetup, we discussed having each team member in the machine learning group claim an algorithm. We've made lot's of progress on the example notebook (`1.TCGA-MLexample.ipynb`) since then (see #18 & #25). Currently, [`1.TCGA-MLexample.ipynb`](https://github.com/cognoma/machine-learning/blob/5eadf0309a409b08fc17d81b36b9c4bcd7acb916/1.TCGA-MLexample.ipynb) uses elastic net logistic regression implemented in `SGDClassifier`.

The goal of this repository is for people to:
1. Claim an algorithm. See the list of classifiers at https://github.com/cognoma/machine-learning/issues/5#issuecomment-235069679. The main requirement is that the algorithm uses the sklearn API so we can use it in the pipeline. Make a comment here once you've chosen an algorithm.
2. Create a modified version of `1.TCGA-MLexample.ipynb` in an `algorithms` directory. So if I took the SVM classifier, I would copy `1.TCGA-MLexample.ipynb` to `algorithms/SVC-dhimmel.ipynb`. Then I would make my edits to `algorithms/SVC-dhimmel.ipynb` to switch to an [SVC](http://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html#sklearn.svm.SVC) classifier.
3. Your goal should be to pick a good set of parameters for grid search. It would also be great if you could document what seems to work well about the algorithm (or if it doesn't seem to work well).

Best of luck! If you can work on this before the [August 9 meetup](https://www.meetup.com/DataPhilly/events/233070705/) then great! Otherwise make sure to bring a laptop with the `cognoma-machine-learning` environment installed.

dhimmel:
Tagging everyone who said they were interested in contributing to the machine learning part of the project in the [introduction thread](https://github.com/cognoma/cognoma/issues/2): @htcai, @loucru1 @bmcgeehan @umeshiso @swbiggs4 @danrieman @rramyr @Inquisitive-Geek @brankaj @yl565 @Ramaa-Nathan @ejsegall  @FadiAlnabolsi @sameertipnis @VijYadav @ctipnis.

Update: also tagging @yigalron.

htcai:
Thanks for the sample notebook! I would like to implement **Linear SVM with regularization**.

dhimmel:
[...]

Issue #22, opened 2016-08-01
Median absolute deviation feature selection


dhimmel:
@gwaygenomics presented evidence that median absolute deviation ([MAD](https://en.wikipedia.org/wiki/Median_absolute_deviation)) feature selection (selecting genes with the highest MADs) can eliminate most features without hurting performance: https://github.com/cognoma/machine-learning/pull/18#issuecomment-236265506. In fact, it appears that performance increased with the feature selection, which could make sense if the selection enriched for predictive features, increasing the signal-to-noise ratio.

Therefore, I think we should investigate this method of feature selection further. Specifically, I'm curious whether:
- @gwaygenomics' findings hold true for outcomes other than RAS?
- MAD is better than MAD / median? I think MAD could be biased against selecting genes that are lowly expressed but still variable?
- MAD outperforms random selection of the same feature set size?
- MAD performs well for other algorithms besides logistic regression?

I'm labeling this issue a task, so please investigate if you feel inclined.

dhimmel:
In 34225ccdefa191287ca153fc14c73bb4eaa6706d -- an example for classifying TP53 mutation -- we did not apply MAD feature selection ([notebook](https://github.com/cognoma/machine-learning/blob/34225ccdefa191287ca153fc14c73bb4eaa6706d/1.TCGA-MLexample.ipynb)). In a8ae61147897aed4a3883853563b357644cbc5f3 (pull request #25), @yl565 selected the top 500 MAD genes ([notebook](https://github.com/cognoma/machine-learning/blob/a8ae61147897aed4a3883853563b357644cbc5f3/1.TCGA-MLexample.ipynb)).

Before MAD feature selection, training AUROC was 95.9% and testing AUROC was 93.5%. After MAD feature selection, training AUROC was 89.9% and testing AUROC was 87.9%. @yl565, did anything else change in your pull request that would negatively affect performance? If not I think we may have an example of 500-MAD genes being detrimental. See @gwaygenomics's [analysis](https://github.com/cognoma/machine-learning/pull/18#issuecomment-236265506) for benchmarking on RAS mutations: 500 genes appears to be borderline dangerous.

yl565:
[...]

PR #28, opened 2016-08-08
Create a template and directory for algorithms

dhimmel:
See #27.

@VijYadav does this pull request answer your questions?


VijYadav:
Daniel- Thank you. Yes, it does. I am working through it.

algorithms/README.md
```diff
@@ -0,0 +1,17 @@
+# Notebooks implementing various sklearn classifiers
+
+This directory hosts notebooks created in response to [Issue #27](https://github.com/cognoma/machine-learning/issues/27#issuecomment-238255454 "cognoma/machine-learning: Claim an sklearn algorithm to implement and troubleshoot") where team members can claim and implement an algorithm.
+
+The recommended approach is to copy `SGDClassifier-template.ipynb` to a file named `algorithm-username.ipynb` and modify the newly created file. Implementations are encouraged to use grid search to find optimal algorithm hyperparameters. Finally, make sure you're using the `cognoma-machine-learning` conda environment described in the [README](https://github.com/cognoma/machine-learning/tree/5eadf0309a409b08fc17d81b36b9c4bcd7acb916#environment).
```
gwaybio: maybe specifying what the actual file format with an example would be good?

```
... a file name "<algorithm>-<username>.ipynb" (e.g. "logistic_regression-gwaygenomics.ipynb")...
```

algorithms/scripts/SGDClassifier-master.py
```diff
@@ -0,0 +1,309 @@
...
+# In[3]:
+
+# We're going to be building a 'TP53' classifier 
+GENE = 'TP53'
+
+
+# In[4]:
+
+# Parameter Sweep for Hyperparameters
+n_feature_kept = 500
```
gwaybio, line 44: We are sticking with 500 here? This is for faster implementations I assume? For a better performance comparison across algorithms I would recommend uping this to at least 1000

gwaybio:
looks good to me @dhimmel - once you address the two comments i will squash and merge

algorithms/scripts/SGDClassifier-master.py
dhimmel, line 44: Agreed although I'm worried people will have [already begun](https://github.com/cognoma/machine-learning/issues/27) with the 500 feature version, so I'm hesitant to switch.

dhimmel, line 44: @gwaygenomics make a decision here and I'll do it.

gwaybio, line 44: ok - leave it as is. We'll need to keep this in mind when we're comparing algorithms however. (I suspect that certain algorithms will perform better or worse with different numbers of features)

gwaybio:
:+1: code and instructions look good to me!

algorithms/scripts/SGDClassifier-master.py
yl565, line 44: @gwaygenomics I agree with you. For performance comparison, I think we should use 8000 (see the graph [here](https://github.com/cognoma/machine-learning/issues/22#issuecomment-238113032)) or may be just keep all features and let everyone pick a combo of feature selection/classification (e.g.. recursive feature elimination + SVM). Though I'm not familiar with gene classification, intuitively feature selection would be crucial here consider we have 20000+ features but only 7000+ samples so I think feature selection should be included as a part of the algorithms  to be tested

George-Zipperlen: I did a git pull today on my laptop, and updated the local machine-learning directory with no problem.  I couldn't find the file SGDClassifier-templater.ipynb mentioned in machine-learning/algorithms/README.md, but found SGDClassifier-master.ipynb 

So I thought to document this...

I edited machine-learning/algorithms/README.md to make this change (I hope that is correct)
then did
        git add README.md
    git commit -m 's/SGDClassifier-template.ipynb/SGDClassifier-master.ipynb'
When I tried 'git push', I got a 403 error.  
Do I also have to do a pull request up in 'https://github.com/cognoma/machine-learning/' ? 
Sorry to bother you with such trivialities.

dhimmel: @George-Zipperlen once you've made your changes you'll submit a pull request, just like you did in the sandbox: https://github.com/cognoma/sandbox/pull/16. Not sure about the 403 error. If you have more issues can you [open a new issue](https://github.com/cognoma/machine-learning/issues/new) rather than commenting here, since this discussion is supposed to be about line 44 of the code above (:

PR #30, opened 2016-08-10
LinearSVC

htcai:
I manipulated two hyperparameters of LinearSVC: **C** and **intercept_scaling**. Both range from 10^-5 to 10. The latter overall does not seem to have much effect on the result. In contrast, an optimal value of **C** is 0.001. Maybe a better model can be found by searching around **C** = 0.001.

The last graph deviates from that in the example notebook. But I think it is probably normal for an SVM. Still, I am not sure how to evaluate the quality of the prediction. Should I add a cell to get the F1 score on the test data?


yl565:
Nice work! I don't think you need to worry about the last graph since SVM do not natively support probability estimation. Instead, you may add a confusion matrix to show the performance of your classifier.

yl565:
You may also try increasing the number of features to be selected or even grid search the optimal n_feature_kept

dhimmel:
Congrats on your first `cognoma/machine-learning` pull request!

I noticed you edited `1.TCGA-MLexample.ipynb` and `algorithms/SGDClassifier-master.ipynb`. You'll want to undo these changes by checking out those files at version 9cfd5cd87025b5ef99dc063e184a3a30eb6d8174 before we merge this pull request.

> The last graph deviates from that in the example notebook. But I think it is probably normal for an SVM.

Good to see that the SVM probabilities don't seem meaningful.

> Still, I am not sure how to evaluate the quality of the prediction.

I would look at the testing AUROC printed in the ROC curve figure legend.

> Should I add a cell to get the F1 score on the test data?

If you think the F1 score would be helpful then feel free to add it. I generally don't find binary predictions very useful, so I stick with performance measures that don't require a specific classification threshold. However, for certain applications the threshold is relevant.

dhimmel:
Make sure to export your notebook to a script (see [this README](https://github.com/cognoma/machine-learning/blob/master/algorithms/README.md)).

htcai:
@dhimmel @yl565 Thank you for your prompt feedback and suggestions! A quick question: should I keep the SGDClassifier notebook in the _algorithms_ directory? I deleted it yesterday. If needed, I can paste it back.

I reviewed the Andrew Ng's Coursera videos and CS229 notes. I can see that SVMs essentially do not aim to provide a probabilistic prediction. I will try what you suggested to evaluate the performance of the model.

yl565:
Before you commit, you should only run `git add` on the files you want to submit, and leave the other files alone so that when you pull request you won't be attempting to change other files. 

Somehow I wasn't able to comment on your code, but I see that in [66]:

``` python
predict_df = pd.DataFrame.from_items([
    ('sample_id', X.index),
    ('testing', X.index.isin(X_test.index).astype(int)),
    ('status', y),
    ('decision_function', pipeline.decision_function(X)),
    ('probability', pipeline.predict(X)),
])
```

You used `pipeline.predict(X)` which output predicted label. This is why you are getting the weird last figure. LinearSVC is based on liblinear which do not support probability output. You could try work on the libsvm based sklearn.svm.SVC which allows you to have probability output at the cost of additional computation time.  I'm not sure if outputting probability could be worthwhile from biology point of view though.

htcai:
@yl565 I have restored the two example notebooks to the initial commit in my local directory.

`predict()` seems to be the only method of prediction for `LinearSVC`. I will try `SVC` and send another pull request after revising `LinearSVC`.

yl565:
It seems that you are still trying to change the 1.TCGA-MLexample.ipynb and algorithms/SGDClassifier-master.ipynb. I suggest you reset your fork to original (see [here](http://stackoverflow.com/questions/9646167/clean-up-a-fork-and-restart-it-from-the-upstream) for instruction) and start commit your file from scrath

dhimmel:
This pull request has been migrated into #33.

Issue #31, opened 2016-08-11
ML module API


maxdml:
Hello,

In an effort to build the global cognoma architecture, it would be very useful to determine an API which defines exactly what is given to the ML module (and incidentally what it will return).

As an exemple of strong API documentation, I believe [OpenStack](http://developer.openstack.org/api-ref.html) is a good start. Note how every module's API is listed, and how for each of those modules [each route is described](http://developer.openstack.org/api-ref-blockstorage-v2.html).

Some direct example for a cognoma API can be found [here](https://github.com/cognoma/django-cognoma/pull/19). This is a first specification for the frontend module.

awm33:
What is the ML module? I don't think the machine learning code will provide any APIs, it will be consuming them. From what it sounds like, the backend group will need to figure out what the input to some function like `run` will be. I think we need to write something that will feed off the task queue and run a function, perhaps loading a class/module specific to an algorithm.

I think the input to this function could be the full [classifier object](https://github.com/cognoma/django-cognoma/blob/master/doc/api.md#get-a-classifier). It would expect a JSON serializable object or map in response.

`results = svg_classifer.run(task.data)`

or even

`svg_classifer.run(task)`'

Where task could even have functions like `task.progress(80, "Fold 4/5 completed")`

maxdml:
The ML module is any library used to implement machine learning algorithm. Actually, if I am not mistaken, this library is [pandas](http://pandas.pydata.org/) (please confirm/correct :).

It will receive request from the Asynchronous Task Queue, as you mentioned. However from the (outdated) [schema](https://github.com/cognoma/cognoma#implementation-details) in the meta cognoma repository, and in the schema @cgreene draw last night (if someone could upload it somewhere that would be awesome), we see the need to decompose the ML component to the django backend.

Even if we just forward the full classifier object, each component has specific APIs, which explains why I opened the issue.

cgreene:
@maxdml : I think the machine learning group is primarily using sklearn, though others may use something else. I think that we should define what gets provided to these methods, and each one should get passed the same information. Maybe some of the implementers can let us know what type of information they use. We should also define what we want the algorithm to report at the conclusion of the run.

awm33:
I think this is more of, what is the schema of the input/output?

[...]

PR #32, opened 2016-08-15
Create LassoCV-brankaj.py

brankaj:
This file contains Python implementation of the Lasso model. This is my first time using pull requests so I am not sure am I doing this right. Any feedback would be great!


dhimmel:
Can you add `algorithms/LassoCV-brankaj.ipynb` to this pull request?

brankaj:
I have added the file, but when I open it on github I cannot see the comments that I have made. Before each modified section I have added a comment. Maybe things should be ok if users open it in Jupyter Notebook.

dhimmel:
The comments do show up [on nbviewer](http://nbviewer.jupyter.org/github/brankaj/machine-learning/blob/f505adf651292db44f638787913d235ac52aa5e3/algorithms/scripts/LassoCV-brankaj.ipynb). Note that GitHub does not render everything in notebooks for security reasons.

To create these comments, it looks like you used a markdown cell but then wrapped it between `<FONT COLOR=\"FF00FF\">\n` and `</FONT>`? Is that the case?

dhimmel:
> In TCGA-MLexample.ipynb two methods are used when presenting output results: decision_function and predict_proba. Lasso model does not have method predict_proba, hence it is removed from the code. Additionally, we obtain a warning that decision_function is going to be deprecated in future versions.

I think the issue is that LassoCV is performing regression rather than logistic regression. In other words, it's not designed for classification. Rather it's designed for predicting a continuous outcome variable. Nevertheless, this pull request is valuable and the method seems to work, just that it's not "conventional" _per se_.

dhimmel:
Everything looks good to me. I'm going to merge unless @brankaj has any changes she would like to make first.

brankaj:
It's great that comments show up on nbviewer. I was searching for an option to change color in markdown cells directly, but the only thing I could find was using HTML tags. The only issue with this approach is that the color attribute is not supported in HTML5 and I would need different code for this.

Anyways, I do not plan to make any more changes in the current file. Thanks for pointing out the continuity requirement for LASSO. There are better ways to incorporate sparsity if we want it. Actually, since in other thread #31 you mentioned that 3 algorithms including LASSO would be implemented using SGDClassifier, I can play with the main code and see what I get when I use l1 penalty in the parameter list of the SGDClassifier. I should get similar results and report this here. 

Thanks for your feedback!

dhimmel:
Merged. Congrats on your first pull request for Project Cognoma. 

> I can play with the main code and see what I get when I use l1 penalty in the parameter list of the SGDClassifier. I should get similar results and report this here.

Would be interesting to see!

brankaj:
Thank you. I plan to report new results by tomorrow evening.

brankaj:
Sorry for the late report, I was doing some other simulations on my computer. Since this thread is closed I am not going to submit new files, just to report scores from the heatmap (obtained from the section Process Mutation Matrix in the main code). 

Lasso model can be obtained using Elastic net with l1_ratio=1. If the regularization term is fixed (i.e., penalty is l1) we can change loss functions. Options for loss function are listed in http://scikit-learn.org/stable/modules/sgd.html#mathematical-formulation. I did not observe significant difference in the scores when using different cost functions (around 84.7%). Only when I increased the number of features to 1000 I got 87.1%. This is still lower than the score obtained when using additional regularization term with 500 features.

PR #33, opened 2016-08-15
LinearSVC: added more evaluation metrics

htcai:
Sorry that it took me a couple of days before I re-sent the pull request. I spent some time getting familiar with some mathematical details about the optimization of SVMs.

I mainly was manipulating two hyper-parameters: the penalty parameter `C` and the scaling of intercept `intercept_scaling`. Both ranges from 1e-5 to 10. The former reaches the optimal results at 0.01 whereas the latter does not seem to have much effect.

The testing AUROC appears to be fine (89.2%), comparing with the training AUROC (92.0%).

Since the prediction of SVMs is always in {0, 1}, it might not be as informative as those models providing probabilistic predictions, especially when cognoma users are querying a single case. For the same reason, I removed the last cell which plots the probabilistic distribution of predictions. However, I added a confusion matrix for the testing data and calculated the F1 score. I think F1 score may also serve as an indication of the confidence of predicting.

I would like to explore other kernel methods if you guys think it is worth a try.

### DIFF
--- a/algorithms/scripts/LinearSVC-htcai.py
+++ b/algorithms/scripts/LinearSVC-htcai.py
@@ -1,0 +1,323 @@
+
+# coding: utf-8
+
+# # Create a logistic regression model to predict TP53 mutation from gene expression data in TCGA
+
+# In[1]:
+
+import os
+import urllib
+import random
+import warnings
+
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+from sklearn import preprocessing, grid_search
+from sklearn.linear_model import SGDClassifier
+from sklearn.svm import LinearSVC
+from sklearn.cross_validation import train_test_split
+from sklearn.metrics import roc_auc_score, roc_curve
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from sklearn.feature_selection import SelectKBest
+from statsmodels.robust.scale import mad
+
+
+# In[2]:
+
+get_ipython().magic('matplotlib inline')
+plt.style.use('seaborn-notebook')
+
+
+# ## Specify model configuration
+
+# In[3]:
+
+# We're going to be building a 'TP53' classifier 
+GENE = 'TP53'
+
+
+# In[4]:
+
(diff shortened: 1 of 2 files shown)


algorithms/scripts/LinearSVC-htcai.py
```diff
@@ -0,0 +1,323 @@
...
+warnings.filterwarnings('ignore', 'using a non-integer number instead of an integer')
+
+ax = sns.distplot(predict_df.query("status == 0").decision_function, hist=False, label='Negatives', color = 'red')
+ax = sns.distplot(predict_df.query("status == 1").decision_function, hist=False, label='Positives', color = 'green')
+
+
+# In[30]:
+
+from sklearn.metrics import confusion_matrix
+from tabulate import tabulate
```
dhimmel: Can you use pandas instead to make a dataframe from `conf_mat`? In short, I don't think we should create a `tabulate` dependency for a trivial task where pandas already provides the functionality.

dhimmel:
Besides the `tabulate` import, everything looks good to me.

> I would like to explore other kernel methods if you guys think it is worth a try.

Sure, why don't we merge this pull request first, and then you can submit an additional one which varies the kernel. I think that could be really valuable.

htcai:
@dhimmel Thanks for your suggestion! That is very reasonable. I have revised the cell. Also, I am happy to try other kernels, which are very interesting to me.

I was struggling with pushing my latest commit for some time. It appeared that the data files are included in the commits, whereas I cannot find it in my local directory. Therefore, I reset the local repo again and then pasted my revised notebook back to the algorithms folder.

dhimmel:
Okay, if you do a `push --force` with the new (clean) repo to your `LinearSVC` branch, the changes will show up in this pull request.

htcai:
Yes, I have already forced the push with my clean local repo. I can see the latest commit in this pull request.

dhimmel:
Cool & congrats on the pull request. Will merge now.

htcai:
Thanks for your feedback and suggestions! I have already learnt a lot by reading and manipulating the notebook.

PR #37, opened 2016-08-24
BernoulliRBM with LogisticRegressionCV

yl565:
Explored feature learning with RBM. The pipeline is as follows:
1. Select n_feat number of features with highest MAD.
2. Scale the n_feat features to within 0-1 range.
3. Fitting a Restricted Boltzmann Machine (RBM) with n_feat visible units and n_RBM_cmp hidden units.
4. Train a logistic regression classifier with the output of RBM as input features.

At the bottom of the notebook, the classification performance with and without RBM is compared (with varying number of input features and RBM hidden units). Seems the investigated feature learning approach doesn't help with improving classification

@gwaygenomics These are the results from what I mentioned during tonight's meetup though I modified the feature scaling to be unsupervised.


algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
```diff
@@ -0,0 +1,199 @@
...
+    """Evaluate and plot the ROC curve of fitted_model
+    """
+    y_pred_train = fitted_model.decision_function(X_train)
+    y_pred_test = fitted_model.decision_function(X_test)
+
+    def get_threshold_metrics(y_true, y_pred):
+        roc_columns = ['fpr', 'tpr', 'threshold']
+        roc_items = zip(roc_columns, roc_curve(y_true, y_pred))
+        roc_df = pd.DataFrame.from_items(roc_items)
+        auroc = roc_auc_score(y_true, y_pred)
```
gwaybio, line 142: should be

``` python
auroc = roc_auc_score(y_true, y_pred, average='weighted')
```

The [default](http://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html) does not account for class imbalance and our classes are generally heavily imbalanced

dhimmel: @gwaygenomics not sure I understand. AUROC is balance-agnostic.

gwaybio: @dhimmel this point is discussed well [here](https://github.com/greenelab/deep-review/issues/43). Setting `average='weighted'` at least takes the weighted average of classification performance for each label

dhimmel: Documentation is confusion... Looks like others [have this confusion](https://www.kaggle.com/c/santander-customer-satisfaction/forums/t/19784/several-ways-to-compute-roc-auc-score/113028).

Here's the source: [`roc_auc_score`](https://github.com/scikit-learn/scikit-learn/blob/234d25677d2929e058f339661a27c8d22fb47bf6/sklearn/metrics/ranking.py#L187) and [`_average_binary_score`](https://github.com/scikit-learn/scikit-learn/blob/234d25677d2929e058f339661a27c8d22fb47bf6/sklearn/metrics/base.py#L32)

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
```diff
@@ -0,0 +1,199 @@
...
+
+# In[4]:
+
+# Fit the model (the computationally intensive part)
+n_features = [500, 2000, 5000]
+n_hiddens = [None, 200, 500, 1000]
+results={'n_feature': [], 'n_RBM_hidden': [], 'train_AUROC': [], 'test_AUROC': []}
+for n_f in n_features:
+    for n_h in n_hiddens:
+        if n_h is not None and n_f < n_h:
```
gwaybio, line 174: I am not too familar with RBMs, they will not work with a hidden layer larger than number of features? Autoencoders can work in this scenario

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
gwaybio: yeah, I have starred at that source code longer than I'd like to...still confusing.

gwaybio:
@yl565 awesome work here! I was not aware that sklean had an [RBM implementation](http://scikit-learn.org/stable/modules/generated/sklearn.neural_network.BernoulliRBM.html#sklearn.neural_network.BernoulliRBM). 

I am surprised by two things that came out of this analysis.
1. How quickly the RBM was trained
   - Did you try increasing the `n_iter` parameter? Maybe you will get better performance if it is trained longer?
2. How well it actually performed
   - I am not surprised however that it performed consistently _worse_ than logistic regression alone

Another thing that I have wrestled with is how to interpret the hidden layers. In an RBM they are either 1 or 0 (so a gene can be on or off). What does this end up pulling out? Pathways? 

Overall, I think this is an important contribution demonstrating that feature engineering using unsupervised learning does work (albeit not as well as using raw features... for now). But we could explore this more maybe through deep belief networks or autoencoders.

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
dhimmel: I did a little test:

``` python
import random
from sklearn.metrics import roc_auc_score

y_true = [0] * 80 + [1] * 20
y_pred = [random.random() for x in range(100)]

options = [None, 'micro', 'macro', 'samples', 'weighted']
for average in options:
    auroc = roc_auc_score(y_true, y_pred, average=average)
    print(auroc)
```

All the `average` options return the same value. I'm pretty sure it doesn't matter for our setup.

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
yl565: I don't work with RBM very often either but I tend to think of RBM as a dimension reduction technique so I figured it's probably not worth producing more features than the raw ones especially consider our sample size. Could there be any case when it is beneficial for the RBM have more outputs than inputs? It would be very interesting.

yl565:
@gwaygenomics Thanks for your comments,  it's exciting to work on a real data set that unsupervised neural net could be useful! My usual data sets are too small/noisy for this.

> 1. How quickly the RBM was trained.
>    - Did you try increasing the n_iter parameter? Maybe you will get better performance if it is trained longer?

I tried n_iter = 5, 15, 30. It seems that in some cases n_iter=5 performs better. I guess overfitting could be one reason behind this. I've [read](https://www.cs.toronto.edu/~hinton/absps/guideTR.pdf) that monitoring the free energy gap between training and validation set could be helpful but it is not yet supported by sklearn. As for the RBM training time, I used only 500-5000 features so I think it is reasonable.

> Another thing that I have wrestled with is how to interpret the hidden layers. In an RBM they are either 1 or 0 (so a gene can be on or off). What does this end up pulling out? Pathways?

It sounds like an interesting thing to explore. I'm not familiar with gene expression but I could print out the trained RBM for you to take a look. 

> But we could explore this more maybe through deep belief networks or autoencoders.

Seems there are people submitted PR on sklearn for [autoencoders and other stuffs](https://github.com/scikit-learn/scikit-learn/issues/3050) few years ago but not finished. We may need to implement them if use only sklearn (looks like a good opportunity to PR in sklearn!)

Finally, for better organization, I think we should add a feature engineering folder for scripts working on feature selection/learning/transform, etc. What's your opinion?

gwaybio:
It would be great to get more functionality embedded in sklearn - expanding there benefits the larger community for sure!

> Finally, for better organization, I think we should add a feature engineering folder for scripts working on feature selection/learning/transform, etc. What's your opinion?

I am all for it! What we're currently exploring in this ML repo are primarily _research_ questions. Eventually the repo will need to communicate with the other cognoma components (frontend, backend, data). But I do think inter-repo communication should be separate from the research questions. This being said, I think the research aspect will be fundamental to how we construct the data and chose algorithms to output results. Also, in thinking about publishing cognoma down the road, it would be great to have also answered an important PanCancer related question.

> I'm not familiar with gene expression but I could print out the trained RBM for you to take a look.

No worries, I will take a look once the repo is merged.

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
gwaybio: I tend to think of techniques like RBM and DAs more as non-linear feature constructors than as dimensionality reduction approaches. Much like you can either interpret a PCA as a dimensionality reduction technique alongside it as a feature extractor. One difference between PCA and RBM/DA is that you can specify _more_ hidden neurons than input. I've never actually seen an instance when this was helpful though.

More info for describing [best practices](http://www.faqs.org/faqs/ai-faq/neural-nets/part3/section-10.html)  of choosing hidden layers.
In general, the "Rules of Thumb" we hear about when picking parameters are [context dependent](http://www.faqs.org/faqs/ai-faq/neural-nets/part4/section-1.html) (See "The Worst" subsection)

gwaybio:
looks good to merge - @dhimmel what do you think?

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
```diff
@@ -0,0 +1,199 @@
...
+            continue
+        if n_h is not None:
+            title = 'n_feature = {0}, RBM hidden units = {1}'.format(n_f, n_h)
+        else:
+            title = 'n_feature = {0}, no RBM applied'.format(n_f)
+        print('\x1b[31m=========================================================================\x1b[0m')
```
dhimmel, line 180: What's going on with this print statement? Python 3 supports unicode, so if you want to include symbols, you should just paste them into this string rather than this approach. Also use string functions to create repetitive strings rather than hardcoding it.

algorithms/scripts/BernoulliRBM-LogisticRegressionCV-yl565.py
```diff
@@ -0,0 +1,199 @@
...
+        print('\x1b[31m            ' + title + ' \x1b[0m')
```
dhimmel, line 181: `'{:^30}'.format('centered')`

https://docs.python.org/3.5/library/string.html

dhimmel:
I don't know anything about BernoulliRBMs, but it seems like if this is a feature reduction method, you should not do it on top of MAD selection. I see the benefit of comparing to MAD selection, but not stacking the two methods in a single pipeline.

dhimmel:
@yl565 looks like this PR is almost good to go. Want to address my two inline comments so we can merge?

yl565:
I want to add a few more analysis when I have time

yl565:
I'm closing this PR. 
P.S. I decided to finish the sklearn [GaussianBernoulliRBM](https://github.com/scikit-learn/scikit-learn/pull/7916) which seems fit the case better.

PR #39, opened 2016-08-26
Machine Learning Pathway Classifier Example

gwaybio:
The pull request adds an example of predicting a Biological Pathway signature using PanCancer data. In the example, I predict Hippo Signaling with variable prediction across tissues. The pull request makes two contributions:
1. Demonstrates that we _can_ predict whole pathways using the Cognoma approach
2. Drives home the need for tissue specific visualizations and the importance of selecting appropriate tissues for training the classifier.


gwaybio:
tagging @stephenshank here too as the pull request is closely related to cognoma/cancer-data#21

gwaybio:
also tagging @allaway - as a cancer biologist, he may have some insight into why predicting Hippo signaling works well in some tissues, but not in others and may also have insight into interpreting some of the top genes.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+# In[4]:
+
+REACTOMEPATHWAY = 'PC7_7459'  # "Signaling by Hippo" in Reactome
+GOPATHWAY_A = 'GO:0035329'    # "hippo signaling" in the Gene Ontology
+GOPATHWAY_B = 'GO:0035330'    # "regulation of hippo signaling" in the Gene Ontology
+
+
+# In[5]:
+
+def pathway_query(path_id, node_type='BiologicalProcess', particip='GpBP'):
```
dhimmel: Can use Neo4j/Cypher parameters and avoid having to do any formatting in python

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+get_ipython().magic('matplotlib inline')
+plt.style.use('seaborn-notebook')
+
+
+# ## Specify model configuration
+
+# In[4]:
+
+REACTOMEPATHWAY = 'PC7_7459'  # "Signaling by Hippo" in Reactome
```
dhimmel: Hardcoded variable names not good. Flexible variables better.

For example, you make a list called `query_params`, each element of the list is a `(identifier, rel_type, node_label)` tuple.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+# In[16]:
+
+# Subset the Y matrix to only the genes to be classified
+y_full = Y[classifier_genes_df['entrez_gene_id']]
+
+
+# In[17]:
+
+y_full.columns = classifier_genes_df['gene_symbol']
+y_full['disease'] = clinical['disease']
```
dhimmel: Can you fix the `SettingWithCopyWarning`? Make a copy of whatever before setting values on slice

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+# In[18]:
```
dhimmel: Doc needed for what this image is: what do the colors mean?

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+mut_tis_total = y_full.groupby('disease').sum()
+mut_tis_total.plot(kind='bar', stacked=True, grid=False, legend=False, sort_columns=True)
+
+
+# In[19]:
```
dhimmel: Right now this shows counts? I think it could be more informative if scaled by total number of samples per disease.

dhimmel: > Specifically, it appears that cancers with gender specificity like testicular cancer and uterine cancer are not impacted.

We want to be sure this isn't just because these cancers have fewer samples.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+# Determine which tissues to filter
+filter_tissue = pd.DataFrame([y, y_full['disease']], index=['sample', 'disease']).T
+tissue_decision = filter_tissue.groupby('disease').sum()
+
+
+# In[21]:
+
+# How many samples in each tissue that have Hippo signaling aberrations
+tissue_decision
```
dhimmel: Would be nice here to have 2 columns (n_positives and n_negatives)

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+# 14. Uterine corpus endometrioid carcinoma
+# 15. Uveal melanoma
+
+# In[23]:
+
+# Distribution of mutation counts after filtering
+sns.heatmap(filtered_mut_total, annot=True)
+
+
+# In[24]:
```
dhimmel: Add comment what this cell is outputting

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+clinical = clinical[clinical['disease'].isin(filtered_mut_total.index)]
+clinical.disease.value_counts()
+
+
+# In[25]:
```
dhimmel: More columns would be nice. In notebook, when displaying full tables, might as well use the space. So like n_pos and n_neg would be helpful.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+# In[37]:
+
+# Cross-validated performance distribution
+facet_grid = sns.factorplot(x='l1_ratio', y='score', col='alpha',
+    data=cv_score_df, kind='violin', size=4, aspect=1)
+facet_grid.set_ylabels('AUROC');
+
+
+# In[38]:
```
dhimmel: This plot begs the question: alpha = 5 or alpha = 10

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+    plt.show()
+
+
+# In[43]:
+
+tissue_results = pd.DataFrame(tissue_auroc, index=['Train', 'Test']).T
+tissue_results = tissue_results.sort_values('Test', ascending=False)
+
+
+# In[44]:
```
dhimmel: `<matplotlib.text.Text at 0x7f613946d0f0>` <-- such output prevents publication grade notebook. Perhaps fix by assigning to variable or including semicolon at end.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+# In[47]:
+
+coef_df.head(10)
+
+
+# The results are very interesting. First, only 192 genes are used to build a fairly successful classifier. Biologists like sparsity! Second, the genes that fall out at the top are informative:
+# 
+# | Entrez | Hugo | Comments |
```
dhimmel: Second column is Symbol not HUGO. These gene symbols are the ones used by Entrez Gene, which will be the same as HUGO probably, but not guaranteed.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+with driver.session() as session:
+    for query in queries:
+        result = session.run(query)
+        result_df = pd.DataFrame((x.values() for x in result), columns=result.keys())
+        full_results_df = full_results_df.append(result_df, ignore_index=True)
+classifier_genes_df = full_results_df.drop_duplicates().sort_values('gene_symbol').reset_index(drop=True)
+classifier_genes_df['entrez_gene_id'] = classifier_genes_df['entrez_gene_id'].astype('str')
+
+
+# In[8]:
```
dhimmel: Perhaps not necessary to print out full table in final version.. maybe just number of genes, especially since figure shows them.

scripts/2.TCGA-MLexample_Pathway.py
stephenshank: Just in case, does anyone else think the identifier we are using here does not match what is currently on Reactome? http://www.reactome.org/content/detail/R-HSA-2028269
Of course we are querying the hetnet. But when I type that ID into Reactome, I don't get anything. Whereas typing the description yields the link above, with a seemingly different ID. I only mention this in case there are potential downstream issues, like our issues with Entrez IDs in the cancer-data group. I may simply be misusing Reactome and this probably doesn't matter, so apologies in advance if this is the case.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+    'https://ndownloader.figshare.com/files/5864859':
+        os.path.join('data', 'expression-matrix.tsv.bz2'),
+    # Y Matrix
+    'https://ndownloader.figshare.com/files/5864859':
+        os.path.join('data', 'mutation-matrix.tsv.bz2'),
+    # Clinical matrix
+    'https://ndownloader.figshare.com/files/5864859':
+        os.path.join('data', 'samples.tsv')
+}
+
```
stephenshank: I think there is a copy/paste error with the URLs here. I tried running this, and it only downloaded a single file. It seems to be because the URLs (the keys of this dictionary) do not change (there is only one), and so only one key gets entered in the final dictionary that gets processed.

scripts/2.TCGA-MLexample_Pathway.py
dhimmel: @stephenshank I think this is because Reactome gene sets where extracted though [Pathway Commons](http://www.pathwaycommons.org/). You're right that attribution to the corresponding Reactome pathway is suboptimal in Hetionet v1.0. Will keep this in mind for the future.

scripts/2.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,535 @@
...
+y.head(5)
+
+
+# ## Set aside 10% of the data for testing
+
+# In[29]:
+
+# Make sure the splits have equal tissue partitions
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0,
+                                                    stratify=clinical['disease'])
```
yl565: I think `y` should also be stratified considering sample size could be small for some disease type

yl565: Seems some disease type also have very imbalanced mutated/not mutated ratio,  together with small sample size you may get very few mutated testing samples with the random split

yl565:
Very interesting problem. Some thoughts here:
If I understand right, there are multiple genes involved here that may be mutated but the classification model do not differentiate between them. Would it be better if we build one model for each gene and combine the models in the decision fusion stage? We may also do a multivariate analysis to group genes with similar expression-mutation interaction (i.e. similar joint distribution **p(X, y)**) and build a model for each group.

yl565:
It seems to me the disease type could also be an important factor to consider. I mean **p(X, y | disease)** could be different depending on **disease**.  Including (predicted) disease in the model may further improve classification performance

dhimmel:
> Would it be better if we build one model for each gene and combine the models in the decision fusion stage?

Interesting, I've never thought about a decision fusion stage. I think one assumption behind building the pathway model is that mutations anywhere in the pathway will produce a cohesive effect. In some ways, it may help you identify this cohesive effect to have all mutations grouped into a single outcome class.

> It seems to me the disease type could also be an important factor to consider.

Yeah, we probably should include disease type as a covariate. Will we have to do something like [`get_dumies`](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.get_dummies.html) or [`OneHotEncoder`](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html#sklearn.preprocessing.OneHotEncoder)? Note that this topic would probably be best discussed in its own issue.

yl565:
@dhimmel It seems to me this is a Multi-label classification problem. We have gene-expression **X** and we want to predict from **X** the disease type and the mutation probability for each gene in the pathway (i.e. a vector of **y** for each single sample in **X**). We can then fuse the predictions in a decision making stage for whatever results the cancer experts may be interested. I haven't worked with a multi-label classification problem before but this seems to be an interesting thing to try. I'm not sure if the approach could be of interest from biology point of view though. @gwaygenomics What do you think?

cgreene:
From my perspective, it's not really clear if this is a multi-label classification problem. If we think that the mutations are all doing something called 'phenocopying' each other (essentially: having more or less the same effect), then really this isn't multi-label and modeling it as such could compromise performance. Greg talked about some examples of this with NF1/Ras in the first cognoma gathering. On the other hand, if they do have distinct effects, then multi-label could be the way to go.

In summary - I don't think there's a clear answer. Particularly if you are using a multi-label approach, then one that allows much of the structure of the solution to be transferred between solutions is probably critical if you're focusing on modeling the effect in one pathway. Because the tissue-background is so different and the number of observations is low, this could compromise predictions a bit - but this is definitely a research question 👍

dhimmel:
Just so we're all clear, here is the definition of multilabel classification [in sklearn](http://scikit-learn.org/0.17/modules/multiclass.html):

> Multilabel classification assigns to each sample a set of target labels. This can be thought as predicting properties of a data-point that are not mutually exclusive, such as topics that are relevant for a document. A text might be about any of religion, politics, finance or education at the same time or none of these.

I agree with @cgreene that multilabel may not be what we want. The motivation for using gene sets to construct the outcome is to increase the number of positives. The multilabel approach will suffer from the low prevalence of mutation for most genes.

However, @yl565 definitely feel free to give multilabel a try if you would like to play around.

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: what do you mean?

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: this is simply an example showing how we can use queries to obtain gene lists. I will update the code but the documentation describing each ID will be important for someone reading the example

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: while I agree it doesn't look pretty to have the genes printed here, it is the one of the two most important aspects a cancer biologist would care about in the classifier (tissue specific performance being the other). They are printed here so someone from that side can quickly scan the genes early on in the classifier script

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: nice catch, I'll fix in the next commit

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: will do!

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: colors represent different mutations - which is hard to display effectively in a legend for this particular plot (since there are 35 genes to display). Do you have any recommendations here? I am leaning towards deleting this plot and adding an extra column to the seaborn heatmap of total samples with hippo mutations

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: > Right now this shows counts?

Yes, it shows counts of total mutations. A sample can contribute only once per column, but can contribute to multiple columns as well. 

> We want to be sure this isn't just because these cancers have fewer samples.

I agree - it is possible the plot can be misleading. I will update in the next commit to reflect a percentage

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: yeah, that would be good

scripts/2.TCGA-MLexample_Pathway.py
dhimmel: > I will update the code but the documentation describing each ID will be important for someone reading the example

The doc can stay as comments:

``` python
[
    ('GO:0035329', 'BiologicalProcess', 'PARTICIPATES_GpBP'), # Doc here AKA "hippo signaling" in the Gene Ontology
    ('GO:0035330', 'BiologicalProcess', 'PARTICIPATES_GpBP'),
    ('PC7_7459', 'Pathway', 'PARTICIPATES_GpPW'),
]
```

scripts/2.TCGA-MLexample_Pathway.py
dhimmel: https://neo4j.com/docs/developer-manual/3.0/cypher/#cypher-parameters

May be a bit difficult with putting the rel_type into a parameter. will help you if you would like.

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: Addressing [this comment](https://github.com/cognoma/machine-learning/pull/39/files#r76476481) removes the need for this input

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: [Same here](https://github.com/cognoma/machine-learning/pull/39/files#r77008683)

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: > Seems some disease type also have very imbalanced mutated/not mutated ratio, together with small sample size you may get very few mutated testing samples with the random split

Nice point, i will change in the next commit

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: alpha in 10 *\* range(-6, 1)?

gwaybio: oh, I see what you're saying...we are not comprehensive in hyperparameter search space... we can explore more of this space in the cognoma implementation

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: got it, thanks!

scripts/2.TCGA-MLexample_Pathway.py
gwaybio: Sounds good!

gwaybio:
@yl565 @dhimmel @cgreene I'm jumping into this discussion a bit late - but I like how we're approaching these research questions. 

My naive guess would be that a classifier built on individual tumors with some sort of fusion model would perform slightly _better_ compared to a pan cancer classifier (but also worse than an ensemble pan cancer classifer!). Of course, this depends on how the fusion is done. Could the fusion model be some sort of weighted vote of individual tissue classifiers applied to all other tissues?

However, like @dhimmel mentioned, these types of conversations are best placed in new issues with some sort of label or milestone called something like: `Standing Research Question`.

scripts/2.TCGA-MLexample_Pathway.py
dhimmel: With the new heatmap of percents and the added columns for the cell titled "# How many samples in each tissue that have Hippo signaling aberrations", this plot isn't providing any unique information. Agree with removal.

scripts/3.TCGA-MLexample_Pathway.py
```diff
@@ -0,0 +1,513 @@
...
+filter_count = 15
+
+tissue_prop_decision = tissue_summary_df['Positive Prop'] >= filter_prop
+tissue_count_decision = tissue_summary_df['Positives'] >= filter_count
+
+tissue_decision = tissue_prop_decision & tissue_count_decision
+tissue_decision
+
+
+# This filters out:
```
dhimmel: This list seems redundant with the output and update-problematic (hardcoded). Would it make sense to just print out this list of filtered tissues?

dhimmel:
Besides one last comment on hardcoded tissue list, LGTM. **Squash** merge at will.

scripts/3.TCGA-MLexample_Pathway.py
gwaybio: Yeah, I think that would be better as well - I'll update now

PR #41, opened 2016-08-27
K Nearest Neighbor

yigalron:
K nearest neighbor - 2 fit methods: one seems to be overfitting, the other one seems better


dhimmel:
Nice work. Can you export your notebook to a .py file for commenting? You can do this by running the following command from the `algorithms` directory:

``` sh
jupyter nbconvert --to=script --FilesWriter.build_directory=scripts KNeighbor-yigalron.ipynb
```

dhimmel:
@yigalron would love to see your work merged into master, so this a friendly nudge :smiling_imp:

yigalron:
Thanks, will do before next meeting; I was away on vacation for the last 2
weeks...

Thanks!
Yigal

yigalron:
@dhimmel ok, ready for review

algorithms/scripts/KNeighbor-yigalron.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+# In[44]:
+
+get_ipython().run_cell_magic('time', '', '# Fit the model (the computationally intensive part)\npipeline.fit(X=X_train, y=y_train)\nbest_clf = clf_grid.best_estimator_\nfeature_mask = feature_select.get_support()  # Get a boolean array indicating the selected features')
+
+
+# In[47]:
+
+get_ipython().run_cell_magic('time', '', "y_pred_train = pipeline.predict(X_train)\ny_pred_test = pipeline.predict(X_test)\n\ndef get_threshold_metrics(y_true, y_pred):\n    roc_columns = ['fpr', 'tpr', 'threshold']\n    roc_items = zip(roc_columns, roc_curve(y_true, y_pred))\n    roc_df = pd.DataFrame.from_items(roc_items)\n    auroc = roc_auc_score(y_true, y_pred)\n    return {'auroc': auroc, 'roc_df': roc_df}\n\nmetrics_train = get_threshold_metrics(y_train, y_pred_train)\nmetrics_test = get_threshold_metrics(y_test, y_pred_test)")
```
dhimmel: Consider predicting probabilities rather than a binary class. This will work better for the ROC curves below.

For example, change to

``` python
y_pred_train = clf.predict_proba(X_train)[:, 1]
y_pred_test = clf.predict_proba(X_test)[:, 1]
```

algorithms/scripts/KNeighbor-yigalron.py
```diff
@@ -0,0 +1,209 @@
...
+plt.ylim([0.0, 1.05])
+plt.xlabel('False Positive Rate')
+plt.ylabel('True Positive Rate')
+plt.title('Predicting TP53 mutation from gene expression (ROC curves)')
+plt.legend(loc='lower right');
+
+
+# In[43]:
+
+# Include loss='log' in param_grid doesn't work with pipeline somehow
```
dhimmel: The `param_grid` above is only relevant for `SGDClassifier`s. It [looks like](http://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html) `KNeighborsClassifier` has some possible parameters that you could evaluate such as `n_neighbors` and `algorithm`.

algorithms/scripts/KNeighbor-yigalron.py
```diff
@@ -0,0 +1,209 @@
...
+from sklearn import neighbors
+from sklearn.neighbors import NearestNeighbors
+from statsmodels.robust.scale import mad
+
+
+# In[42]:
+
+n_neighbors = 5
+n_feature_kept = 500
+param_fixed = {
```
dhimmel: This can be removed since it's not used.

dhimmel (changes requested):
Congrats on the pull request.

So it looks like the difference between the first and second is that the first figure corresponds to a classifier built directly on `X_train` whereas the second uses the pipeline to apply additional operations before fitting the classifier?

I'd recommend, only using the pipeline to simplify the notebook. If you don't want to do one of the transformations, just remove it from the pipeline.

It would also be nice to specify some options for `param_grid` based on KNeighborsClassifier.

algorithms/scripts/KNeighbor-yigalron.py
yigalron: should I implement these changes on the pull request I already have, or do I need to create a new pull?

dhimmel: Implement the changes in this pull request. You can do so by adding more commits to your `master` branch.

yigalron: OK, thanks; I can't make the meeting tonight. I implemented some of the changes already and I'll keep working on it in the next few days. I noticed that the choice of 5 neighbors is actually good: if I select more or fewer I get worse results.

yigalron:
I committed a new set of changes; did you get a notification? basically I removed some unused code and also tried different settings for number of neighbors; I also tried "uniform" vs. "distance" classification, and found that the uniform works better.

dhimmel:
@yigalron, yep got the notification. I'm aiming to get around to reviewing this today.

KNeighbor-yigalron.py
```diff
@@ -0,0 +1,173 @@
...
+
+clf = neighbors.KNeighborsClassifier(n_neighbors,'uniform')
+# clf = neighbors.KNeighborsClassifier(n_neighbors,'distance')
+
+# joblib is used to cross-validate in parallel by setting `n_jobs=-1` in GridSearchCV
+# Supress joblib warning. See https://github.com/scikit-learn/scikit-learn/issues/6370
+warnings.filterwarnings('ignore', message='Changing the shape of non-C contiguous array')
+clf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')
+pipeline = make_pipeline(
+    feature_select,  # Feature selection
```
dhimmel: I think we should remove the MAD feature selection. We don't really need to save time, since you're code is running quickly. I think it could help performance to have more features to determine "nearness".

KNeighbor-yigalron.py
```diff
@@ -0,0 +1,173 @@
...
+from sklearn import neighbors
+from sklearn.neighbors import NearestNeighbors
+from statsmodels.robust.scale import mad
+
+
+# In[2]:
+
+n_neighbors = 5
+n_feature_kept = 500
+param_grid = {
```
dhimmel: You should put your parameters in this `param_grid`.

For example:

``` python
param_grid = {
    'n_neighbors': [5, 10, 20],
    'weights': ['uniform'],
}
```

dhimmel (changes requested):
Nice work, the notebook mostly looking good.

> I removed some unused code

:+1: always a good idea!

> tried different settings for number of neighbors; I also tried "uniform" vs. "distance" classification, and found that the uniform works better.

The grid search is designed to make it easy to try different parameter combinations in a disciplined way. So the recommended approach is to put add every setting you want to try to `param_grid`. Can you modify this pull request so you put the different parameter options in `param_grid` rather than manually trying them?

dhimmel:
@yigalron, great work getting this pull request up to speed.

It looks like somewhere your notebook became duplicated. `KNeighbor-yigalron.ipynb` and `KNeighbor-yigalron.ipynb` exist in the repository's root directory as well as in `algorithms`. We'll want make sure the final notebooks from this pull request are in `algorithms` and there are no new files in the root directory.

The results are looking really good -- strong performance and nice ROC curves. There isn't over-fitting in training. It would be great to print out the optimal parameters chosen by grid search.

In the current notebook at fdbf984b518f7ee118098631d02a597e73fec8b6, we don't know which value for `n_neighbors` was deemed optimal. You can retrieve this with `clf_grid.best_params_`. Can you add this output? Sorry if you have to run the whole thing again!

dhimmel:
@yigalron, this is great. So 20 nearest neighbors gave the best cross-validated performance. Good to know. One final thing that would be helpful would be to see is the mean AUROC for each parameter combination. You should be able to retrieve this info from `clf_grid` after you fit the model.

That's all I can think of -- then I think we'll be ready to merge.

KNeighbor-yigalron.py
```diff
@@ -126,22 +127,48 @@
...
+            row['score'] = score
+            rows.append(row)
+    df = pd.DataFrame(rows)
+    return df
+
+
+# In[24]:
+
+cv_score_df = grid_scores_to_df(clf_grid.grid_scores_)
+facet_grid = sns.factorplot(x='n_neighbors', y='score', col='n_neighbors',
```
dhimmel: `n_neighbors` is used as both the x variable and column variable. I would remove `col='n_neighbors'`, which will compact the plot. You could also set `col='weights'` -- however this would only make a difference if you were evaluating multiple options for `weights`, which is not currently the case.

Otherwise, this is a beautiful and very informative visualization.

KNeighbor-yigalron.py
```diff
@@ -0,0 +1,190 @@
+
```
dhimmel: This file should be moved to `algorithms/scripts/KNeighbor-yigalron.py`.

I think you can do `git mv KNeighbor-yigalron.py algorithms/scripts/KNeighbor-yigalron.py`. Also you'll need to move `KNeighbor-yigalron.ipynb` to `algorithms/KNeighbor-yigalron.ipynb`.

dhimmel (changes requested):
The cross-validation visualization is really helpful.

Just two small issues remaining.

dhimmel:
The cross-validation plot of performance for various number of neighbors in [this notebook](https://github.com/cognoma/machine-learning/blob/3c05ff8f7518e2213c615c3d97281b1a85633e06/algorithms/KNeighbor-yigalron.ipynb) looks great!

Only remaining item is to remove the two files in the root directory.

yigalron:
done - please check it out.

dhimmel (approved):
Great work seeing this PR through and contributing the K Nearest Neighbors to Cognoma!

yigalron:
Thanks! hopefully i'll be able to get to the next meeting and see how we
move this forward.

PR #42, opened 2016-08-31
Automate cancer-data download from figshare

dhimmel:
`1.download.ipynb` queries the figshare API to find the latest versions of the files and downloads them.

Changes directory for storing downloaded data from `data` to `download`, following the convention of `cognoma/cancer-data`.

Add `requests` library to environment.

Add an execute script that executes notebooks in the root directory in order and exports them to scripts.

In `algorithms`, notebooks were not updated besides `SGDClassifier-master.ipynb` which is the example notebook and template for new implementations.


dhimmel:
@gwaygenomics, based on whether #39 or this is merged first, one of us will have to do some extra work.

algorithms/scripts/SGDClassifier-master.py
```diff
@@ -35,7 +35,7 @@
 # In[3]:
 
 # We're going to be building a 'TP53' classifier 
-GENE = 'TP53'
+GENE = '7157' # TP53
```
gwaybio, line 38: Is `algorithms/scripts/SGDClassifier-master.py` the nbconverted `2.TCGA-MLexample.ipynb`? It is difficult to compare the diffs between `1.TCGA-MLexample.ipynb` and `2.TCGA-MLexample.ipynb`

scripts/1.download.py
```diff
@@ -0,0 +1,52 @@
+
+# coding: utf-8
```
gwaybio, line 2: I think this is an important script and should be the number 1 file in the repo - but I did like the fact that both example scripts could run independently. It looks like with this update a user will need to first run the download script

scripts/1.download.py
```diff
@@ -0,0 +1,52 @@
...
+    os.mkdir('download')
+
+
+# In[6]:
+
+# Download the files specified by the metadata
+for file_info in response['files']:
+    url = file_info['download_url']
+    name = file_info['name']
+    print('Downloading {} to `{}`'.format( url, name))
```
gwaybio: extra space here between "format(" and "url"?

scripts/1.download.py
dhimmel: You could make multiple files starting with two, if you wanted to show they were parallel. Not sure whether that will turn out to be a good practice or not

gwaybio: not sure what you mean? I am saying that running either example (either the`TP53` or `Hippo`) directly after cloning would output the result without having to first download the data. With this change (and the corresponding change in [lines 69-80 of `algorithms/SGDClassifier-Master.py](https://github.com/cognoma/machine-learning/pull/42/files#diff-085839bf1754a46003c4dfbba1c979f8L69)) the user would need to run `scripts/1.download.py` first. 

Hmmm...I guess this would be ok and moving toward a more succinct workflow would be good. A user would be able to see the ipython notebook anyway with the results embedded. Ok I think I talked myself out - no longer concerned!

algorithms/scripts/SGDClassifier-master.py
dhimmel: I had forgotten to delete `scripts/1.TCGA-MLexample.py`. I did a force push. Now see the diff for "scripts/1.TCGA-MLexample.py → scripts/2.TCGA-MLexample.py"

scripts/1.download.py
dhimmel: Does my latest commit clear up confusion. Analysis notebooks will no longer have any download logic?

gwaybio:
LGTM - excited about the progress! I will now remove the download logic in #39

dhimmel:
Okay will merge now. You will probably have to rebase your PR.

Issue #21, opened 2016-07-28
What covariates should we include as features?


dhimmel:
In addition to gene expression, we probably should include other information on samples. This discussion will focus on identifying potential covariates and evaluating whether they make sense to include in models. If we don't include the right covariates, confounding is likely to be an issue.

See #8 as a potential example of confounding that may be addressable by adding a mutation load feature.

dhimmel:
@stephenshank began work at the last meetup on creating a `covariates.tsv` from `samples.tsv`. @stephenshank any updates?

stephenshank:
My first attempt at this can be found at #46, where I simply try to process the samples data to begin using these as features. Immediate issues are...
1. Are we comfortable with the way categorical NaN's are handled?
2. How do we want to treat numeric NaN's? The current pipeline breaks when data contains NaN's.
3. Standardizing column names... perhaps all as lowercase/underscore, which seems to be consistent?

As a longer term issue, I was eager to do some actual machine learning, but my naive attempt at using the existing classifier failed. So it would be great to get some discussion going, regarding what tweaks we expect to enhance performance. Now reading through more issues, I realize I should've tried this on some of the Hippo pathway genes. But I am worried that munging the features together as I have is not an effective approach.

Of course I would also welcome discussion on anything I may have missed... @dhimmel, @gwaygenomics any thoughts?

dhimmel:
> Are we comfortable with the way categorical NaN's are handled?

As per my review comment on #46: yes, although imputation is definitely an option here.

> How do we want to treat numeric NaN's? The current pipeline breaks when data contains NaN's.

I know of 3 options: impute, filter observations, or remove variable. Since we don't want to start hemorrhaging samples, I don't think we should filter many observations. So maybe we can assess imputation/removal on a per-variable basis. i.e. will it impute, if so impute and keep, else remove.

> Standardizing column names... perhaps all as lowercase/underscore, which seems to be consistent?

Personally, I make all_lowercase_underscore_separated variable names. However, I do see a benefit in not messing with Xena names unless we store a reversible mapping. In other words, if using foreign data, it's sometimes better to use dirty column names than break interoperability. However, lot's of these variables have already been changed or recoded in cancer-data, so interoperability is less of a worry for those variables.

stephenshank:
> Personally, I make all_lowercase_underscore_separated variable names

[...]

Issue #44, opened 2016-09-12
Decisions required to reach a minimum viable product


dhimmel:
We're nearing the point where we'll need to implement a machine learning module to execute user queries. We're looking to create a minimum viable product. We can expand functionality later, but for now let's focus on the simplest and most succinct implementation. There are several decisions to make:
1.  **Classifier**: which classifiers should we support? If we want to support only a single classifier for now, which one?
2. **Predictions**: do we want to return probabilities, scores, or class predictions?
3. **Threshold**: do we want to report performance measures that depend on a single classification threshold? Or do we want report performance that span thresholds?
4. **Testing**: Do we want to use a testing partition in addition to cross-validation? If so, do we refit a model on all observations?
5. **Features** Should we include covariates in addition to expression features (see #21)?
6. **Feature selection**: Do we want to perform any feature selection?
7. **Feature extraction**: Do we want to perform features extraction, such as PCA (see #43)?

So let's work out these choices, with a focus on simplicity.

dhimmel:
Here are my thoughts:
1.  **Classifier**: `sklearn.linear_model.SGDClassifier` with a grid search to find the optimal `l1_ratio` and `alpha`. See [`2.TCGA-MLexample.ipynb`](https://github.com/cognoma/machine-learning/blob/99bff9268c20a3142f8b46d2a923b2cadf254e12/2.TCGA-MLexample.ipynb) for an example.
2. **Predictions**: let's return all three using the following object names `probability`, `score`, `class` under a `predictions` key. The frontend should handle cases where `probability` is absent. 
3. **Threshold**: Both.
4. **Testing**: Let's hold out 10% for testing.
5. **Features** deferring this decision based on the maturity of #21.
6. **Feature selection**: let's do MAD feature selection to 8000 genes based on @yl565's findings in https://github.com/cognoma/machine-learning/issues/22#issuecomment-238113032. This should help speed up fitting the elastic net without too much performance loss.
7. **Feature extraction**: deferring this decision based on the maturity of #43.

@gwaygenomics, @yl565, @stephenshank: do you agree?

gwaybio:
Can you clarify what you mean by number 3?

> Or do we want report performance that span thresholds?

Like AUROC?

dhimmel:
[...]

PR #46, opened 2016-09-15
First attempt at processing covariate information.

stephenshank:
All feedback is welcome. Also going to get some discussion going in #21.


scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+from sklearn.feature_selection import SelectKBest
+from statsmodels.robust.scale import mad
+get_ipython().magic('matplotlib inline')
+
+
+# Let's peak at the sample data.
+
+# In[2]:
+
+samples_df = pd.read_table('download/samples.tsv', index_col=0)
```
dhimmel: Use `os.path.join` to construct filename for windows compatability.

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+samples_df.isnull().sum(axis=0)
+
+
+# Now count the number of types of each column. We'll use Python's set, and therefore must exclude null values.
+
+# In[4]:
+
+categorical_variables = samples_df.columns[[1, 2, 3, 4, 6, 8]]
+for variable in categorical_variables:
+    not_null = ~samples_df[variable].isnull()
```
dhimmel: Pandas has a `notnull` function

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+print('Total number of samples: %d' % len(samples_df))
+print('Number of nulls in each column:')
+samples_df.isnull().sum(axis=0)
+
+
+# Now count the number of types of each column. We'll use Python's set, and therefore must exclude null values.
+
+# In[4]:
+
+categorical_variables = samples_df.columns[[1, 2, 3, 4, 6, 8]]
```
dhimmel: Best to avoid referring to columns by index, since indexes are unstable across versions and aren't explicit in their meaning. List column names instead?

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+# 
+# Let's save this to a <code>.tsv</code> file.
+
+# In[11]:
+
+path = os.path.join('download', 'covariates.tsv')
+covariates_df.to_csv(path, sep='\t')
+
+
+# Now we'll try to see whether incorporating this information will help to improve the elastic net classifier that was explained at a past meetup.
```
dhimmel: Let's save the remainder of this notebook for another notebook/pull request. You'll be able to really clean up your imports with this gone as well.

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+# Let's process the entire thing. We'll convert the categorical variables to dummies, and then concatenate the numeric variables (age of diagnosis, days survived, and days recurrence free).
+
+# In[10]:
+
+columns_to_process = [samples_df.disease,
+                      samples_df.organ_of_origin,
+                      samples_df.gender,
+                      mortality,
+                      recurred]
+columns_as_dummies = [pd.get_dummies(column) for column in columns_to_process]
```
dhimmel: `get_dumies` can take the whole dataframe at once... No need to do this column by column.

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+        label='{} (AUROC = {:.1%})'.format(label, metric['auroc']))
+plt.xlim([0.0, 1.0])
+plt.ylim([0.0, 1.05])
+plt.xlabel('False Positive Rate')
+plt.ylabel('True Positive Rate')
+plt.title('Predicting TP53 mutation from gene expression (ROC curves)')
+plt.legend(loc='lower right');
+
+
+# Unfortunately, the classifier does not even seem to notice the presence of the covariates. But this could be due to the MAD filtering step, lack of feature normalization, and plenty of other things. Curious to hear thoughts about where to go.
```
dhimmel: My guess is that the MAD selection is removing the covariates. But even if not, covariates can still be helpful even if they don't improve overall performance -- they can still prevent confounding (faulty feature importance measurements).

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+# Inspecting the head of the samples DataFrame above, we see that two columns, namely <code>dead</code> and <code>recurred</code>, need some attention. We're going to encode categorical variables using panda's get_dummies. Since these columns are indicated by a 1 or 0, this will become the column header when encoded, as below.
+
+# In[6]:
+
+pd.get_dummies(samples_df.dead).head(10)
+
+
+# Note that null values are encoded as (0,0) pairs... I was wondering if we were okay with this.
```
dhimmel: I think that's a decent way to handle NaN without having to remove the sample. Obviously not ideal but it works. Another option would be imputation. Maybe a future pull request can work on imputing these variables (especially gender).

scripts/4.Covariates.py
```diff
@@ -0,0 +1,222 @@
...
+# 
+# Let's rename the values in each of these so that they more accurately reflect the underlying variable.
```
dhimmel, line 64: How about rename after converted to dummies, using `DataFrame.rename`?

This way you can rename for all variables at once using one big dictionary.

dhimmel (changes requested):
@stephenshank awesome pull request. I requested some changes. Just to be clear about the desired scope of this pull request: the goal should be creating a `covariates.tsv` from `samples.tsv` (and nothing more).

See my exploratory notebook #46 which encodes some variables as dummies. The processing in #46 isn't intended to become part of the main repository (notebooks in the root directory), but may help you see ways to use `get_dummies`. It also would be nice to add `n_mutations_log10` as in #46. This will be an important covariate.

dhimmel:
In my previous review #46 should be #47. Oops!

scripts/4.Covariates.py
stephenshank: Done.

scripts/4.Covariates.py
stephenshank: Done.

scripts/4.Covariates.py
stephenshank: Very good to know thanks!

scripts/4.Covariates.py
stephenshank: This ties into your comment above about using a big dictionary to rename all the columns. When you use `get_dummies` on the whole data frame, it prepends the name of the column to each categorical variable. I.e., the new columns names for the `recurred` column will be `recurred_has_recurred` and `recurred_has_not_recurred`.

Are we okay with this? Without also formatting the disease/organ names, it starts to look a little strange. For instance, `organ_of_origin_Thyroid Gland` as a column name. Typing out the big dictionary is doable but slightly tedious... perhaps there is room to be clever with Python or regexes.

It's definitely minor, but any advice on how to proceed for the next PR would be appreciated.

scripts/4.Covariates.py
stephenshank: Okay, I suppose I was a bit too eager for this haha.

scripts/4.Covariates.py
```diff
@@ -56,7 +48,7 @@
 
 # In[5]:
 
-categorical_variables = categorical_variables[1:]
+del categorical_variables[categorical_variables.index('sample_type')]
```
dhimmel, line 51: See [`DataFrame.drop`](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.drop.html)

scripts/4.Covariates.py
```diff
@@ -67,156 +59,98 @@
...
+
+# Let's process the entire thing. We'll convert the categorical variables to dummies, and then concatenate the numeric variables (age of diagnosis, days survived, and days recurrence free).
+
+# In[13]:
+
 numeric_columns = ['age_diagnosed', 'days_survived', 'days_recurrence_free']
-covariates_df = pd.concat(columns_as_dummies+[samples_df[numeric_columns]], axis=1)
+mutation_column = np.log1p(number_of_mutations)
+covariates_df = pd.concat(
+    [samples_df[numeric_columns], pd.get_dummies(samples_df[categorical_variables])],
```
dhimmel, line 113: I think you can use the `columns` arg of `get_dummies` so you don't have to mess with concat. Not sure if maybe you have another motive like column order here.

scripts/4.Covariates.py
```diff
@@ -0,0 +1,156 @@
...
+# In[7]:
+
+mortality_map = pd.Series(['alive', 'dead'], index=[0.0, 1.0])
+samples_df.dead = samples_df.dead.map(mortality_map)
+samples_df.dead.head(10)
+
+
+# In[8]:
+
+recurred_map = pd.Series(['has_not_recurred', 'has_recurred'], index=[0.0, 1.0])
```
dhimmel, line 75: Not sure if you understood my previous comment here. The suggestion is to avoid any value mapping here. Then once you create `covariates_df` you can use

``` python
renamer = {
    'dead_0': 'alive',
    'dead_1': 'dead',
   # keep going for everything you want to rename
}
covariates_df.rename(columns=renamer, inplace=True)
```

As an aside, I think it's more succinct to map using a dictionary rather than creating a series.

scripts/4.Covariates.py
```diff
@@ -0,0 +1,156 @@
...
+
+path = os.path.join('download', 'covariates.tsv')
+covariates_df.to_csv(path, sep='\t')
+
+
+# Now some exploratory data visualization for imputing missing numeric values. First, the plot of age versus number of mutations from the previous notebook.
+
+# In[15]:
+
+sns.jointplot('n_mutations_log1p', 'age_diagnosed', data=covariates_df, kind='reg')
```
dhimmel, line 131: I think plot is a duplicate of [cell 9 here](https://github.com/cognoma/cancer-data/blob/67f80326c4514d1b0d4d42a744fd85b971c24f7f/3.explore-mutations.ipynb). Maybe remove this plot but keep the disease specific plots?

dhimmel (changes requested):
Some more comments. Will respond to your get_dummies naming question next.

Also can you track `covariates.tsv` in git? It's not too big right?

scripts/4.Covariates.py
dhimmel: My latest suggestion suggests you rename `recurred_0` to `has_not_recurred`. As far as `organ_of_origin_Thyroid`, I'd leave it --- it's ugly but makes it easy to get back to the Xena organ encoding.

dhimmel: But if you think you have a better solution, suggest.

scripts/4.Covariates.py
stephenshank: Oops... yes it is now clear! Will redo.

scripts/4.Covariates.py
stephenshank: Nevertheless, are you okay with the `concat`? Everything is properly indexed, so I believe it is correct. If it's not okay, it is not immediately clear to me how to use the `columns` argument...

stephenshank: I suppose I could drop the categorical variables once they are converted to dummies, is this what you had in mind?

dhimmel: Not a problem if you would like to keep it, but I think this replaces 112-113:

``` python
covariates_df = pd.get_dummies(samples_df, columns=categorical_variables)
```

stephenshank: But then it does not have the numeric covariates?

stephenshank:
`covariates.tsv` is < 3 MBs, so we should be able to track. If there is any strong preference for where it should reside, let me know, otherwise I will put it at the top level in the next PR.

Edit: On second thought, I will create a covariates directory with this notebook and put the tsv file there.

scripts/4.Covariates.py
dhimmel: > But then it does not have the numeric covariates?

My understanding and experience is that it retains non-categorical columns, just doesn't encode them.

stephenshank: Sorry, this is indeed correct! It is not at all clear from the pandas documentation...

scripts/4.Covariates.py
```diff
@@ -0,0 +1,156 @@
...
+covariates_df['n_mutations_log1p'] = mutation_column
+covariates_df.head()
+
+
+# Finally, let's save this to a <code>.tsv</code> file.
+
+# In[14]:
+
+path = os.path.join('download', 'covariates.tsv')
+covariates_df.to_csv(path, sep='\t')
```
dhimmel, line 124: consider `float_format='%.5g'` to suppress useless decimal points. Should reduce file size perhaps.

dhimmel:
I would suggest perhaps making this notebook 1 in this repo and writing `covariates.tsv` to a `data` directory.

However, I'm starting to think that this work belongs in `cancer-data`? @gwaygenomics and @stephenshank, what do you think about adding it the the notebook pipeline in cancer data?

yl565:
Didn't know if it applies here but I often replace missing value with the predictions from other variables

yl565:
Is this information supposed to be used for mutation classification? If this is the case, perhaps it is better to handle the NaNs at the classification stage since each algorithm may have different preference?

stephenshank:
@dhimmel I was actually thinking the same thing. I would be happy to move this over to `cancer-data` as the latest notebook, with `covariates.tsv` being placed in `data` so that it is tracked. Your suggestion above reduced file size to ~ 1.2 MB. Does this sound okay?

dhimmel:
> Is this information supposed to be used for mutation classification? If this is the case, perhaps it is better to handle the NaNs at the classification stage since each algorithm may have different preference?

@yl565, are there any sklearn models that handle missing values? What do you mean by "handle the NaNs at the classification stage"?

> file size to ~ 1.2 MB. Does this sound okay?

stephenshank, sounds great.

> I would be happy to move this over to cancer-data as the latest notebook, with covariates.tsv being placed in data so that it is tracked.

Okay let's move this to `cancer-data`. You should open a new PR there that references this pull request.

yl565:
@dhimmel  `sklearn.preprocessing.Imputer` handles missing values which can be included in the classification pipeline. I mean for people doing mutation classification, they may prefer to process the missing values themselves rather than working with already pre-processed data (at least I am...). There are different ways to deal with the missing data which could interact with classification algorithms differently.

yl565:
A recommended way to deal with missing value could be set as default and it would be nice to have the option to get the raw, unpreprocessed data

dhimmel:
> There are different ways to deal with the missing data which could interact with classification algorithms differently.

Interesting... how about we perform imputation in `cancer-data` resulting in two possible covariate datasets: `covariates.tsv` and `covariates-imputed.tsv` (which would use a single imputation method we decide on). Then machine learning users would be free to select either the imputed covariates or perform imputation themselves.

yl565:
That would be nice!

gwaybio:
Really nice analysis! I thought the age vs. mutation burden by tissue is very interesting... However, not sure it would be safe to impute age this way. 

In general, my thoughts on imputation are as follows:
1. We should definitely be sure to separate imputed covariates from non-imputed
   - I believe that our target audience (cancer biologists) will not view imputed variables kindly
2. Survival associated covariates (age, vital status, time to recurrence, survival time) are tough to impute and even tougher to convince people we are imputing them correctly
   - [Survival analyses](https://en.wikipedia.org/wiki/Survival_analysis) are typically performed using this data and they have well studied methods for dealing with missing or data
3. Missing data is a common problem for clinical metrics (even more of a problem for electronic health records!) so imputation is more of a research question than a part of our minimum viable product (see #44 )
4. Imputing gender should be easy enough, lets at least try this one!

That being said, it could still be a interesting exercise and research question to try to impute all missing info.

Issue #8, opened 2016-07-20
Could a general mutation-load pattern confound mutation-specific signals?


dhimmel:
I think it's likely that there is a general expression pattern for how mutated a tumor is. For example, super mutated tumors may have wacky gene expression, solely because they're super mutated and not specifically because of which exact mutations they contain.

For a given gene, tumors with mutations are more likely to be highly mutated overall. This could cause confounding. It may appear that a mutation is associated with a specific expression pattern, although the signal is be driven by general mutation-load.

So we may need to end up including a mutation-load covariate. In the meantime, someone should see whether it's possible to use gene expression to predict the mutation-load of each sample (labeling this a task and looking for a volunteer).

cgreene:
This is an interesting question. From some quick searching of the academic literature, I've dug up mutational signatures of heavy mutation load cancers (the types of mutations that occur in these cancers seem to be different). I didn't find anything on a gene expression pattern common to them. It may be important to control for confounding by cancer type (maybe you pick the most mutated 10% within each cancer type as positive and the least mutated 10% as negative). I think this is an interesting question that may have just created another use case!

gwaybio:
on a call now and this issue was mentioned. Really, the issue is mainly that these hyper-mutated tumors have a ton of passenger mutations and would contaminate gold standards. The solution proposed involved subsetting mutations using [Cancer Hotspots](http://cancerhotspots.org/#/home) as defined by [Chang _et al._](http://www.ncbi.nlm.nih.gov/pubmed/26619011)

Essentially what the group is doing is only considering a sample to have a mutation in a given gene if the mutation is found in this database. I don't necessarily know what to do with this info - or if it even makes sense to use at all but generally, using it would increase the percentage of true positives but simultaneously increase false negatives.

dhimmel:
> it would increase the percentage of true positives but simultaneously increase false negatives

What do you mean by true positives and false negatives?

From [Chang et al.](https://doi.org/10.1038/nbt.3391):

> Here, we developed a statistical algorithm to identify recurrently mutated residues in tumor samples. We applied the algorithm to 11,119 human tumors, spanning 41 cancer types, and identified 470 somatic substitution hotspots in 275 genes.

So if we were to only count mutations that were in recurrently mutated residues (cancer hotspots), we would only be able to offer our users a choice between 275 genes — not good? Additionally, I'm not sure I see:
1. that restricting to hotspots will be able to fully eliminate mutation load confounding
[...]

PR #47, opened 2016-09-15
Evaluate performance of covariates at predicting various mutations

dhimmel:
Creates an explore directory and README for this type of exploratory notebook.

See how well covariates (non-expression features) predict TP53 mutation.

Related to https://github.com/cognoma/machine-learning/issues/8: General mutation-load does provide some ability to predict mutation status of TP53.

Partially addresses https://github.com/cognoma/machine-learning/issues/21: Covariates are extracted from `samples.tsv`.


cgreene:
> Related to #8: General mutation-load does provide some ability to predict mutation status of TP53.

This isn't too surprising given TP53's role in controlling cell cycle checkpoints. Is this only true for genes in cell cycle checkpoint or DNA repair pathways, or is it also true for genes in proliferation pathways?

dhimmel:
> Is this only true for genes in cell cycle checkpoint or DNA repair pathways, or is it also true for genes in proliferation pathways?

@cgreene, I updated the notebook to evaluate performance of covariate-only classifiers for the 8 interesting mutations we've [previously considered](https://github.com/cognoma/cancer-data/pull/22#issuecomment-245294155). Here is performance of the models with all covariates included:

covariate-performance

So actually, _TP53_ is among the hardest to predict using only covariates. _VHL_ which is highly disease-specific achieves a near-perfect AUROC. Therefore, without the disease or organ covariate, expression classifiers of _VHL_ are likely just classifying kidney clear cell carcinoma / kidney tissue.

See [this dataframe](https://github.com/dhimmel/machine-learning/blob/50aa17482dd1f9ebfae7d98fcf98a0209eb86487/explore/confounding/auroc.tsv) to get a general idea of covariate importance. Mutation load and disease type both seem important.

cgreene:
@dhimmel : Interesting! I can imagine that gene expression would clearly capture disease/organ, so if one usually had a mutation in a single gene that was relatively specific (e.g. VHL) I could imagine that creating a strong signal.

explore/confounding/confounding.py
```diff
@@ -0,0 +1,280 @@
+
```
gwaybio, line 1: looks like you need to `nbconvert` this again

gwaybio: I'll take a look again once it's easier to read

gwaybio (changes requested):
looks like nbconvert `confounding.ipynb` has to be rerun

explore/confounding/confounding.py
dhimmel: Good catch. should be fixed now.

explore/README.md
```diff
@@ -0,0 +1,9 @@
+# A directory for exploratory machine learning analyses
+
+This directory is home is exploratory analyses that help answer questions about how we should do machine learning. For algorithm implementations see the [`algorithms`](../algorithms) directory. For other types of analyses, place them here.
```
gwaybio: "This directory is home **to** exploratory analyses"?

explore/confounding/confounding.py
```diff
@@ -0,0 +1,193 @@
...
+# coding: utf-8
+
+# # Create a logistic regression model to several mutations from covariates
```
gwaybio: "Create a logistic regression model to **predict** several mutations from covariates"?

explore/confounding/confounding.py
```diff
@@ -0,0 +1,193 @@
...
+
+# In[6]:
+
+def expand_grid(data_dict):
+    """Create a dataframe from every combination of given values."""
+    rows = itertools.product(*data_dict.values())
+    return pd.DataFrame.from_records(rows, columns=data_dict.keys())
+
+mutations = {
+    '7157': 'TP53', # tumor protein p53
```
gwaybio: pep8 says inline comments should be separated by [two spaces](https://www.python.org/dev/peps/pep-0008/#inline-comments)

explore/confounding/confounding.py
```diff
@@ -0,0 +1,193 @@
...
+    'survival': ['alive', 'dead'],
+}
+
+
+# ## Compute performance
+
+# In[8]:
+
+def get_aurocs(X, y, series):
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
```
gwaybio, line 135: what does `series` look like?

gwaybio (changes requested):
Interesting and informative analysis. My comments are generally cosmetic, but it would be nice to get some additional documentation on some of the functions!

gwaybio (approved):

PR #49, opened 2016-09-19
Linear Discriminant Analysis

htcai:
I replaced feature selection with Linear Discriminant Analysis. The feature data was compressed to 2000 dimensions. Also, I removed the plotting of coefficients, since the resultant features do not have intuitive interpretations as do the initial features.

It is my first time to implement LDA in a pipeline. I just added an instance of LDA to the pipeline. I suppose that the trained pipeline will transform the testing data when it is used to make predictions.

I tracked and printed the max memory during the training on my MacBook, which seems to compress memory. As a consequence, the printed max memory is much smaller than what I saw in the Activity Monitor.

There is significant over-fitting. The scores for training and testing are 99.4% vs. 90.4%.

### DIFF
--- a/algorithms/scripts/SGDClassifier-LDA-htcai.py
+++ b/algorithms/scripts/SGDClassifier-LDA-htcai.py
@@ -1,0 +1,254 @@
+
+# coding: utf-8
+
+# # Create a logistic regression model to predict TP53 mutation from gene expression data in TCGA
+
+# In[1]:
+
+import os
+import urllib
+import random
+import warnings
+import resource
+
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+from sklearn import preprocessing, grid_search
+from sklearn.linear_model import SGDClassifier
+from sklearn.cross_validation import train_test_split
+from sklearn.metrics import roc_auc_score, roc_curve
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from sklearn.discriminant_analysis import LinearDiscriminantAnalysis
+from sklearn.decomposition import PCA
+from statsmodels.robust.scale import mad
+
+
+# In[2]:
+
+get_ipython().magic('matplotlib inline')
+plt.style.use('seaborn-notebook')
+
+
+# ## Specify model configuration
+
+# In[3]:
+
+# We're going to be building a 'TP53' classifier 
+GENE = '7157' # TP53
+
+
+# In[4]:
+
+# Parameter Sweep for Hyperparameters
+n_feature_pca = 300
+n_feature_kept = 10
+param_fixed = {
+    'loss': 'log',
+    'penalty': 'elasticnet',
+}
+param_grid = {
+    'alpha': [10 ** x for x in range(-5, 2)],
+    'l1_ratio': [0, 0.05, 0.1, 0.2, 0.5, 0.8, 0.9, 0.95, 1],
+}
+
+
+# *Here is some [documentation](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html) regarding the classifier and hyperparameters*
+# 
+# *Here is some [information](https://ghr.nlm.nih.gov/gene/TP53) about TP53*
+
+# ## Load Data
+
+# In[5]:
+
+get_ipython().run_cell_magic('time', '', "path = os.path.join('..', 'download', 'expression-matrix.tsv.bz2')\nX = pd.read_table(path, index_col=0)")
+
+
+# In[6]:
+
+get_ipython().run_cell_magic('time', '', "path = os.path.join('..', 'download', 'mutation-matrix.tsv.bz2')\nY = pd.read_table(path, index_col=0)")
+
+
+# In[7]:
+
+y = Y[GENE]
+
+
+# In[8]:
+
+# The Series now holds TP53 Mutation Status for each Sample
+y.head(6)
+
+
+# In[9]:
+
+# Here are the percentage of tumors with NF1
+y.value_counts(True)
+
+
+# ## Set aside 10% of the data for testing
+
+# In[10]:
+
+# Typically, this can only be done where the number of mutations is large enough
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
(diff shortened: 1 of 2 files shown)


algorithms/scripts/SGDClassifier-LDA-htcai.py
```diff
@@ -0,0 +1,243 @@
...
+# In[21]:
+
+predict_df = pd.DataFrame.from_items([
+    ('sample_id', X.index),
+    ('testing', X.index.isin(X_test.index).astype(int)),
+    ('status', y),
+    ('decision_function', pipeline.decision_function(X)),
+    ('probability', pipeline.predict_proba(X)[:, 1]),
+])
+predict_df['probability_str'] = predict_df['probability'].apply('{:.1%}'.format)
```
gwaybio, line 232: Do you get an error or warning with this assignment? If you do, you could use

``` python
predict_df = predict_df.assign(probability_str=predict_df['probability'].apply('{:.1%}'.format))
```

gwaybio (commented):
Very informative analysis - I just had one question. Also, I will wait on merging until you decide if you wanted to include your analysis last night of different `num_features_kept`

algorithms/scripts/SGDClassifier-LDA-htcai.py
htcai: Thanks for your questions! Firstly, changing the n_feature_kept to 500 or 100 does not reduce overfitting, as I found last night. I have also tried to search over larger values of alpha [0.01, 0.1, 1, 10, 100, 1000], but the larger values led to much worse performance in cross-validation. Therefore, the optimal alpha is 0.01 and there is no noticeable change in testing. I am looking for solutions.

By the way, there is no warning or error at the code of predict_df.

dhimmel (changes requested):
`download-data.ipynb` duplicates functionality with `1.download.ipynb`. Also these algorithm pull requests shouldn't really be adding notebooks to the root directory.

What is the difference between `feature-transformation.ipynb` in root and `algorithms/algorithms/SGDClassifier-LDA-htcai.ipynb`?

htcai:
Sorry for the unknown files. I am not sure how they were produced. But I have reset my local repo to the earlier commit. I am trying to see whether I can reduce the over-fitting of LDA. The pull request will be updated soon.

htcai:
Sorry for the delay. I have searched for solutions and experimented different strategies. Finally, it turns out that applying PCA before LDA solves the problem of overfitting, following [this suggestion](http://stats.stackexchange.com/questions/106121/does-it-make-sense-to-combine-pca-and-lda).

Specifically, PCA maps the initial data onto a 300-dimensional space and then LDA reduces the dimensionality to 10. Comparing with the [approach of using only PCA to produce 2000 features](https://github.com/cognoma/machine-learning/blob/master/algorithms/SGDClassifier-PCA-mans2singh.ipynb), my LDA-PCA hybrid method achieves a higher testing AUROC (93.8% vs. 93.1%). In addition, the training is also faster (total time 10min 29s vs. 17min 50s).

It might be interesting to perform a grid search over the two parameters for PCA and LDA (while keeping the hyper-parameters of the SGDClassifier constant. But the training definitely will take much more time.

dhimmel:
Very nice, can you export `algorithms/SGDClassifier-LDA-htcai.ipynb` to a script (`.py`) file. From `algorithms` run:

``` sh
# Export notebook to a .py script for diff viewing
jupyter nbconvert --to=script --FilesWriter.build_directory=scripts SGDClassifier-LDA-htcai.ipynb
```

dhimmel:
@htcai here's one question. It looks like [`LinearDiscriminantAnalysis`](http://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html#sklearn.discriminant_analysis.LinearDiscriminantAnalysis) in sklearn can be used as a classifier. For example, it has a predict method. However, if I'm not mistaken, in your example it's used only as a transformer, since it's not the final step of the pipeline. Am I thinking about this correctly?

htcai:
Hi Daniel, thank you for reminding me of generating the py file! It has been added to the scripts directory.

In addition, I had the same question regarding incorporating `LinearDiscriminantAnalysis` into the pipeline, and I understand the documentation of [Pipeline](http://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html) in the same way as you do. In order to examine the correctness of this usage of `LinearDiscriminantAnalysis`, I have created a separate branch in my repo, in which I decomposed the pipeline into separate steps. The [output](https://github.com/htcai/machine-learning/blob/LDA-nopipeline/algorithms/SGDClassifier-LDA-htcai.ipynb) is the same as that of using `Pipeline`, though I eliminated the section **Investigate the predictions**.

dhimmel (approved):

algorithms/scripts/SGDClassifier-LDA-htcai.py
```diff
@@ -0,0 +1,254 @@
...
+
+# Include loss='log' in param_grid doesn't work with pipeline somehow
+clf = SGDClassifier(random_state=0, class_weight='balanced',
+                    loss=param_fixed['loss'], penalty=param_fixed['penalty'])
+
+# joblib is used to cross-validate in parallel by setting `n_jobs=-1` in GridSearchCV
+# Supress joblib warning. See https://github.com/scikit-learn/scikit-learn/issues/6370
+warnings.filterwarnings('ignore', message='Changing the shape of non-C contiguous array')
+clf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')
+pipeline = make_pipeline(
```
dhimmel, line 120: It turns out our current pipeline setup is flawed (see https://github.com/cognoma/machine-learning/pull/54) which could make your cross-validated accuracy appear higher. Testing should still be accurate. No need to fix in this pull request. Will merge.

htcai:
Thanks for merging the pull request! I am also curious about the (minor) flaw in the pipeline.

In addition, I am wondering whether we should specify the `stratify` parameter in `train_test_split`, since the proportions of positive and negative samples are far from each other. Using the stratified splitting, the frequencies of positive samples will be approximately the same in the training and the testing data.

dhimmel:
> I am wondering whether we should specify the stratify parameter in train_test_split

Yes, we should stratify by status in `train_test_split`. `cognoml` will do this https://github.com/cognoma/machine-learning/pull/51/files#diff-4607d4a968832a9c9b0c00b2624655c0R56.

PR #40, opened 2016-08-27
Finished AdaBoost classifier

beelze-b:
AUROC of .867 on test set

Highest weighted 3 were: KLRG2, METTL7B, and GPR158

Using a decision stump classifier as the weak learner for AdaBoost, the best parameters were:
a learning rate of  0.5 and 50 estimators.

More estimators seem to perform best, and more hyperparameter tuning can be done.

PR #50, opened 2016-09-20
Adding Factor Analysis

beelze-b:
Finally got factor analysis to work. 

I was trying to generate all the components to full dimensions and then reducing to the most important components, but the memory constraints were just too large. I tried just using 5000 components, but that took over 2 hours to run. 

Like Yichuan suggested in his code, I just ended up doing 500 components.
Although not too rigorous, it works, and it still provides a boost in performance.

The best hyper parameters for the SGDclassifier were {'alpha': 0.01, 'l1_ratio': 0.1}
with Factor Analysis.

The AUC went from .904 to .939 on the test set.

Runtime was around 20 minutes with about 9 GB of memory used (including all my RAM and some swap space).


dhimmel:
Can you remove `algorithms/AdaBoost-NabeelSarwar.ipynb` from this pull request? You may have something weird going on since both this pull request and your previous (https://github.com/cognoma/machine-learning/pull/40) were based on your master branch.

Other than that, looks good to me.

beelze-b:
Hi, I removed adaboost from the pull request. It should work fine.

dhimmel:
Nice, one last thing. Can you export `FactorAnalysis-NabeelSarwar.ipynb` to a script by running the following from the `algorithms` directory:

``` sh
jupyter nbconvert --to=script --FilesWriter.build_directory=scripts FactorAnalysis-NabeelSarwar.ipynb
```

beelze-b:
Script is added. Also i won't be able to make tonight as I am waiting for a delivery at home. I can work remotely if I know what is new.

dhimmel (approved):

PR #51, opened 2016-09-20
Create the cognoml package to implement an MVP API

dhimmel:
I'll remove WIP (work in progress) when this pull request is complete.

See https://github.com/cognoma/machine-learning/issues/31 for API design discussion.

See https://github.com/cognoma/machine-learning/issues/44 for MVP (minimal viable product) feature inclusion decisions.


cognoml/analysis.py
```diff
@@ -85,6 +86,11 @@ def classify(sample_id, mutation_status, **kwargs):
         performance[part] = utils.value_map(metrics, round, ndigits=5)
     performance['cv'] = {'auroc': round(clf_grid.best_score_, 5)}
     results['performance'] = performance
+    
+    results['model'] = utils.model_info(clf_grid.best_estimator_)
+
+    feature_df = utils.get_feature_df(clf_grid.best_estimator_, X.columns)
```
dhimmel: @yl565 currently I'm retrieving feature names using `X.columns` which will get the feature names of `X` before it enters the pipeline. However, since `VarianceThreshold` or other feature selection/tranformation steps will alter the feature set, do you know how we can get feature names at the end of the pipeline? In other words, we want the feature names corresponding to `clf_grid.best_estimator_.coef_`. I searched for like an hour and couldn't figure this out.

cognoml/analysis.py
```diff
@@ -56,26 +56,34 @@ def classify(sample_id, mutation_status, **kwargs):
...
-    #obs_df['predicted_prob'] = pipeline.predict_proba(X)
-        
-    is_testing = obs_df.testing.astype(bool)
-    y_pred_train = obs_df.predicted_score[~is_testing]
-    y_pred_test = obs_df.predicted_score[is_testing]
+    predict_df = pd.DataFrame.from_items([
+        ('sample_id', X_whole.index),
+        ('predicted_status', pipeline.predict(X_whole)),
+        ('predicted_score', pipeline.decision_function(X_whole)),
+        ('predicted_prob', pipeline.predict_proba(X_whole)[:, 1]),
```
dhimmel: @yl565 what's the best way to see if a pipeline supports `predict_proba`? We can upgrade to sklearn 18 once that's released, if that will make things easier.

dhimmel:
This pull request is ready for some review. Suggesting @yl565 @awm33 @cgreene @gwaygenomics.

The package name is `cognoml`. Feedback on name welcome.

The package is pip installable, but there are outstanding issues:
- **data management**: I'm not sure the best way to have our package access the data. Should you always have to pass a path for the directory that contains the data? If the data doesn't exist in that directory, it can be downloaded.
- **algorithm support**: currently only SGDClassifier is supported. It's not clear how we want to modularly support different pipelines. If we want to support different algos, how should this info get passed to this module?
- [**feature name bug**](https://github.com/cognoma/machine-learning/pull/51#discussion_r79705959)
- currently the returned JSON may contain numpy values which break the builtin JSON encoder. Alternatively, `cognoml.utils.JSONEncoder` should work. However, if writing to the database, maybe we should just sanitize all JSON values before returning.

To run the example, use:

``` sh
python cognoml/main.py > data/api/hippo-output.json
```

cognoml/analysis.py
```diff
@@ -0,0 +1,103 @@
...
+    dimensions['positives'] = sum(obs_df.status == 1)
+    dimensions['negatives'] = sum(obs_df.status == 0)
+    dimensions['positive_prevalence'] = y.mean().round(5)
+    dimensions['training_observations'] = len(obs_train_df)
+    dimensions['testing_observations'] = len(obs_test_df)
+    results['dimensions'] = utils.value_map(dimensions, round, ndigits=5)
+
+    performance = collections.OrderedDict()
+    for part, df in ('training', obs_train_df), ('testing', obs_test_df):
+        y_true=df.status
```
gwaybio: pep8 spacing updates needed

scripts/3.TCGA-MLexample_Pathway.py
```diff
@@ -111,17 +112,17 @@
 
 # In[9]:
 
-get_ipython().run_cell_magic('time', '', "path = os.path.join('data', 'expression-matrix.tsv.bz2')\nX = pd.read_table(path, index_col=0)")
+get_ipython().run_cell_magic('time', '', "path = os.path.join('download', 'expression-matrix.tsv.bz2')\nX = pd.read_table(path, index_col=0)")
```
gwaybio, line 98: if someone were to nbconvert the `ipynb` file will this script be overwritten?

cognoml/analysis.py
```diff
@@ -0,0 +1,103 @@
...
+    obs_df = pd.DataFrame.from_items([
+        ('sample_id', sample_id),
+        ('status', mutation_status),
+    ])
+    
+    X_whole = read_data()
+    X = X_whole.loc[obs_df.sample_id, :]
+    y = obs_df.status
+    
+    X_train, X_test, y_train, y_test = train_test_split(
```
gwaybio, line 56: Confirming that you're deciding not to stratify based on disease too?

cognoml/analysis.py
```diff
@@ -0,0 +1,103 @@
...
+    gs = utils.value_map(gs, utils.df_to_datatables)
+    results['grid_search'] = gs
+    
+    results['model'] = utils.model_info(clf_grid.best_estimator_)
+
+    feature_df = utils.get_feature_df(clf_grid.best_estimator_, X.columns)
+    results['model']['features'] = utils.df_to_datatables(feature_df)
+
+    results['observations'] = utils.df_to_datatables(obs_df)
+    return results
```
gwaybio, line 114: Maybe in the beginning of this script you can describe what `results` should look like? I am having some difficulty interpreting what results actually entails and its format

gwaybio (commented):
Nice work. I only made a couple minor comments. I can also confirm that the example `data/api/hippo-output.json` works as expected!

scripts/3.TCGA-MLexample_Pathway.py
dhimmel: Shouldn't be as the upstream changes to `3.TCGA-MLexample_Pathway.ipynb` are part of this pull request.

gwaybio: ah ok, see it now

cognoml/analysis.py
dhimmel: Ah, stratification by disease could also make sense. Currently, sample/covariate info is not part of this pull request. I think it probably should be added before the first release.

gwaybio: also, I was at talk by Olivier Elemento - he was building models for a different purpose (predict immunotherapy responders) but was adjusting for mutation burden as a covariate. We may want to consider checking out his stuff and adjusting for burden too

cognoml/analysis.py
dhimmel: I generated a [JSON schema](http://json-schema.org/) using ([`hippo-output-schema.json`](https://github.com/dhimmel/machine-learning/blob/28cb22b1f7ec6c3438241e617d1bbc6d94a1410c/data/api/hippo-output-schema.json)):

```
genson --indent=2 data/api/hippo-output.json  > data/api/hippo-output-schema.json
```

I can add descriptions for each field here. Do you think that's a good solution.

gwaybio: Maybe just a link to the hippo-output-schema and the genson command would suffice

gwaybio:
ok, i think this looks good to me. Although you may want to wait on @yl565 for those specific questions.

cognoml/analysis.py
dhimmel: See d52c6a2a60615951e03b0e84eadbe07d718e908e for my solution

cognoml/analysis.py
dhimmel: See https://github.com/scikit-learn/scikit-learn/issues/7536

dhimmel:
I propose we merge this sooner rather than later. The pull request is already quite large.

@gwaygenomics, @cgreene, @yl565 does one of you want to submit a review for up to 2291a0cd0afc74d3888cf04a52a169ca3435de19?

README.md
```diff
@@ -20,3 +20,13 @@ conda env update --file environment.yml
...
+
+## `cognoml` package
+
+This repository contains a python package named `cognoml`. The package can be installed locally by running the following command from this repository's root directory:
+
+```sh
+pip install --editable .
+```
+
+Make sure the `cognoma-machine-learning` environment is activated first, so the `cognoml` is only installed for this environment.
```
gwaybio: Would it make sense to say this requirement before `pip install`?

cognoml/analysis.py
```diff
@@ -0,0 +1,114 @@
...
+from cognoml.figshare import download_files
+
+# expression_path = os.path.join('download', 'mutation-matrix.tsv.bz2')
+data_directory = "download"
+
+def read_data(version=None):
+    """
+    Read data.
+    """
+    v_dir = download_files(directory=data_directory, artile_id=3487685, version=version)
```
gwaybio: typo in `article_id`

also, is there a reason you are not providing `article_id` and `directory` as function arguments?

cognoml/analysis.py
```diff
@@ -0,0 +1,114 @@
...
+        ('sample_id', sample_id),
+        ('status', mutation_status),
+    ])
+    
+    X_whole = read_data(version=data_version)
+    X = X_whole.loc[obs_df.sample_id, :]
+    y = obs_df.status
+    
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=0.1, random_state=0, stratify=y)
```
gwaybio, line 57: again, this is always set to 10%?

cognoml/classifiers/logistic_regression.py
```diff
@@ -0,0 +1,30 @@
...
+
+param_grid = {
+    'classify__random_state': [0],
+    'classify__class_weight': ['balanced'],
+    'classify__loss': ['log'],
+    'classify__penalty': ['elasticnet'],
+    #'classify__alpha': 10.0 ** np.arange(-3, 2),
+    #'classify__l1_ratio': [0.0, 0.1, 0.2, 0.5, 0.8, 0.9, 1.0],
+    'classify__alpha': 10.0 ** np.arange(-1, 1),
+    'classify__l1_ratio': [0.0, 1.0],
```
gwaybio: Lasso or Ridge only?

cognoml/figshare.py
```diff
@@ -0,0 +1,60 @@
...
+def get_article_versions(article_id):
+    """
+    Get version_to_url dictionary for a figshare article.
+    """
+    url = 'https://api.figshare.com/v2/articles/{}/versions'.format(article_id)
+    response = requests.get(url)
+    version_to_url = {d['version']: d['url'] for d in response.json()}
+    return version_to_url
+
+def download_files(directory, artile_id=3487685, version=None):
```
gwaybio: another typo in `article_id`

cognoml/figshare.py
```diff
@@ -0,0 +1,60 @@
...
+        article_id on figshare
+    version : int or None
+        figshare version. `None` means latest.
+    
+    Returns
+    -------
+    str
+        The version-specific DOI corresponding to the downloaded data.
+    """
+    version_to_url = get_article_versions(artile_id)
```
gwaybio: article_id

cognoml/figshare.py
```diff
@@ -0,0 +1,60 @@
...
+    Get version_to_url dictionary for a figshare article.
+    """
+    url = 'https://api.figshare.com/v2/articles/{}/versions'.format(article_id)
+    response = requests.get(url)
+    version_to_url = {d['version']: d['url'] for d in response.json()}
+    return version_to_url
+
+def download_files(directory, artile_id=3487685, version=None):
+    """
+    Download files for a specific figshare article_id and version to the specified directory.
```
gwaybio: a usage note that it won't _really_ download to the specified dictionary - it will append the version too

cognoml/utils.py
```diff
@@ -0,0 +1,132 @@
...
+
+def json_sanitize(obj, object_pairs_hook=collections.OrderedDict):
+    """
+    Sanitize an object containing pandas/numpy objects so it's JSON
+    serializable. Does not preserve order since `pandas.json.dumps()` does not
+    respect OrderedDict objects. Hence, it's recommended to just use the builtin
+    `json.dump` function with `cls=JSONEncoder`.
+    """
+    obj_str = pd.json.dumps(obj)
+    print(obj_str)
```
gwaybio, line 46: making sure you want to print this here

data/api/hippo-output-schema.json
```diff
@@ -0,0 +1,323 @@
+{
```
gwaybio, line 1: I didn't read this file

gwaybio (changes requested):
nice comments and flow - easy to review. I didn't try running this iteration yet (I believe I tested one previously). But I can run if you're not pressed for time.

I only had some minor fixes and general questions.

cognoml/utils.py
dhimmel: Yeah this main function is currently only for running a single example / test case. Actual users will call the `cognoml.analysis.classify()` function.

cognoml/analysis.py
dhimmel: Yeah unless you have another heuristic that you think is better. Eventually we will probably need to smarten up here and potentially refuse to classify problems with less than a certain number of positives.

cognoml/analysis.py
dhimmel: > also, is there a reason you are not providing article_id and directory as function arguments?

What do you mean?

gwaybio: the `read_data()` function has only a single argument: `version` - you have `download_files` here accepting two variables `data_directory` and `3487685` that are blind to the function.

If this was the functionality you are intending, I would recommend doing this instead:

``` python
def read_data(directory=data_directory, article_id=3487685, version=None):
    """
    Read data.
    """
    v_dir = download_files(directory=directory, article_id=article_id, version=version)
    ...
```

cognoml/classifiers/logistic_regression.py
dhimmel: Fixing the pipeline (#54) really slowed things down because the tranformation steps are now performed separately for each CV fold. Therefore, I really cut down the grid. In retrospect, this grid is probably too small -- I'll increase it.

dhimmel: See https://github.com/cognoma/machine-learning/issues/56 and 10308e0f40278409a257b4c6119359a63ca22a98

cognoml/analysis.py
dhimmel: `data_directory` is a module scope variable that should be set using:

``` python
cognoml.analysis.data_directory = 'new_path'
```

Using the default definition of `directory=data_directory` will break the above functionality.

There is no support for changing `article_id` currently.

So while I see your point, I think the changes add more complexity without any additional functionality at the moment.

dhimmel: @gwaygenomics [note that](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments):

> Python’s default arguments are evaluated once when the function is defined, not each time the function is called (like it is in say, Ruby).

gwaybio: For mutable defaults only?

dhimmel: > For mutable defaults only?

No for all defaults. Their value is evaluated upon definition. This leads to an odd behavior for mutable defaults where they can be modified with each function call. Immutables don't have this issue, but are still evaluated upon definition.

gwaybio (approved):

PR #52, opened 2016-09-29
Classifier results for different genes

brankaj:
Hi all,

Since MVP is going to be released soon and open to the public, maybe it would be useful to provide some kind of demo to explain how the package works. It would be like providing tutorial/examples for the first-time users. 
Also, since most likely sklearn.linear_model.SGDClassifier is going to be used as a classifier, it would be nice to see the results for different genes, not only TP53. During the last meeting there was a discussion regarding this. Since it would take really long time to run the code for all genes, maybe it's better to focus on the group of genes that are considered dominant by a certain criteria. 
I have submitted pull requests for two files. Both files output AUROCs for a list of genes. First file is a code that determines AUROCs for a list of tumor suppressor genes. The list is not exhaustive and it is copied from https://www.cancerquest.org/cancer-biology/cancer-genes/tumor-suppressors/tumor-suppressor-table. The second code runs the classifier for the genes that are associated with the lung cancer. This list of genes is obtained from https://ghr.nlm.nih.gov/condition/lung-cancer#genes.

Not sure is this going to be useful to the researchers in the field, but I thought giving it a try. 
All feedback welcome!


algorithms/2.TCGA-MLexample-LungCancer.py
```diff
@@ -0,0 +1,166 @@
...
+# In[5]:
+
+get_ipython().run_cell_magic('time', '', "path = os.path.join('download', 'mutation-matrix.tsv.bz2')\nY = pd.read_table(path, index_col=0)")
+
+
+# In[6]:
+
+# GENES: AKT1, ALK, BRAF, DDR2, EGFR, ERBB2, KRAS, MAP2K1, NRAS, PIK3CA, PTEN, RET, RIT1, ROS1
+
+list_of_tsgenes = [207,238,673,4921,1956,2064,3845,5604,4893,5290,5728,5979,6016,6098]
```
dhimmel: Consider using a dictionary of entrez_gene_id to symbol. This way you can easily make the plot labels use gene symbols which are more recognizable.

algorithms/2.TCGA-MLexample-LungCancer.ipynb
```diff
@@ -0,0 +1,311 @@
...
+    "\n",
+    "    # joblib is used to cross-validate in parallel by setting `n_jobs=-1` in GridSearchCV\n",
+    "    # Supress joblib warning. See https://github.com/scikit-learn/scikit-learn/issues/6370\n",
+    "    warnings.filterwarnings('ignore', message='Changing the shape of non-C contiguous array')\n",
+    "    clf_grid = grid_search.GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')\n",
+    "    pipeline = make_pipeline(\n",
+    "        feature_select,  # Feature selection\n",
+    "        StandardScaler(),  # Feature scaling\n",
+    "        clf_grid)\n",
+    "    %%time\n",
```
dhimmel: I think this `%%time` should be removed -- it doesn't look like it's really working :smile_cat:

dhimmel (changes requested):
Really nice exploration! It was interesting to see how overfitting got worse when the positive prevalence decreased.

Let's move this to the `explore` directory rather than the `algorithms` where it is now.

Also, would it be possible to consolidate this analysis to one notebook. DRY -- do not repeat yourself. Right now almost all the code is repeated. The one notebook can create two plots: a lung cancer gene plot and a suppressor gene plot but most of the code will be shared.

algorithms/2.TCGA-MLexample-LungCancer.py
brankaj: Thank you for the feedback! I used ids listed on the NCBI website. I made sure that the numbers correspond to the Homo sapiens column :bowtie: 
I will prepare one notebook with the incorporated changes. Time to learn about dictionaries. 
Btw, since I do not see the option to move the files to the explore dictionary, is it easier that I close/delete this pull request and start a new one with one notebook?

dhimmel: > Thank you for the feedback! I used ids listed on the NCBI website. I made sure that the numbers correspond to the Homo sapiens column

That's the correct way to do it. See [this example](https://github.com/cognoma/cancer-data/blob/b054721bbde1ed0abc19f395eb7629728870cb7c/scripts/3.explore-mutations.py#L100-L109) for what I mean by the dictionary. The goal is to make your plot labels show gene symbols rather than entrez GeneIDs.

dhimmel:
> Btw, since I do not see the option to move the files to the explore ~~dictionary~~ directory, is it easier that I close/delete this pull request and start a new one with one notebook?

No need to create a new pull request. You can just make the changes to your `brankaj-patch-2` branch. [Here's some info](https://githowto.com/moving_files) on how to move files that are tracked in git.

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,246 @@
...
+
+for GENE in list_of_genes:
+    
+    y = Y[GENE]
+    # Here are the percentage of tumors with NF1
+    per = y.value_counts(True)
+    # Set aside 10% of the data for testing, this can only be done where the number of mutations is large enough
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
+    'Size: {:,} features, {:,} training samples, {:,} testing samples'.format(len(X.columns), len(X_train), len(X_test))
+    def fs_mad(x, y):
```
dhimmel: You can define `fs_mad` and `get_threshold_metrics` outside of the for loop. Currently, every iteration of the loop redefines the function which is unnecessary because the function does not change.

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,246 @@
...
+        roc_columns = ['fpr', 'tpr', 'threshold']
+        roc_items = zip(roc_columns, roc_curve(y_true, y_pred))
+        roc_df = pd.DataFrame.from_items(roc_items)
+        auroc = roc_auc_score(y_true, y_pred)
+        return {'auroc': auroc, 'roc_df': roc_df}
+
+    metrics_train = get_threshold_metrics(y_train, y_pred_train)
+    metrics_test = get_threshold_metrics(y_test, y_pred_test)
+    
+    AUROCS_train += [metrics_train['auroc']]
```
dhimmel: You can use `append` to add a single element to the end of a list:

``` python
AUROCS_train.append(metrics_train['auroc'])
```

dhimmel (changes requested):
Nice analysis. It's encouraging to see that all classifiers had testing performance with AUROC > 50%.

Just two minor comments.

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,252 @@
...
+    auroc = roc_auc_score(y_true, y_pred)
+    return {'auroc': auroc, 'roc_df': roc_df}
+
+
+# In[8]:
+
+for GENE in list_of_genes:
+    
+    y = Y[GENE]
+    # Here are the percentage of tumors with NF1
```
gwaybio: Comment mentioning `NF1` no longer applies - this could be confusing inside the loop over genes

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,252 @@
...
+    
+    AUROCS_train.append(metrics_train['auroc'])
+    AUROCS_test.append(metrics_test['auroc'])
+    p.append(per[1])
+  
+
+
+# In[9]:
+
+AUROCS_train_lung = AUROCS_train[0:len(list(genes_LungCancer.values()))]
```
gwaybio: no need to convert `genes_LungCancer` to a list, this should work fine:

``` python
AUROCS_train_lung = AUROCS_train[0:len(genes_LungCancer)]
```

This change would propagate through lines below as well

gwaybio (commented):
This is a fantastic analysis - It is really nice to show that Cognoma works well for a variety of genes! I really like the bar charts as well...

This may be beyond the scope of the current pull request, but plotting performance for some of the oncogenes would be helpful as well (However, I do see a couple in the lung cancer genes analysis).

brankaj:
Thanks a lot for the comments. The details regarding the latest file are given in the git commit message. Let me know if you prefer that I restore the get_threshold_metrics() function.

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,291 @@
...
+    '7490':'WT1',
+}
+
+genes_Oncogenes = {
+    '5155':'PDGFB', #growth factor
+    '5159':'PDGFRB', #growth factor 
+    '3791':'KDR', #receptor tyrosine kinases
+    '25':'ABL1', #Cytoplasmic tyrosine kinases
+    '6714':'SRC', #Cytoplasmic tyrosine kinases
+    '5894':'Raf-1',#cytoplasmic serine kinases
```
dhimmel: The [primary symbol](https://www.ncbi.nlm.nih.gov/gene/5894) for this gene is RAF1. In general human gene names are all upercase.

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,291 @@
...
+    '1029':'CDKN2A',
+    '1630':'DCC',
+    '4089':'SMAD4',
+    '4087':'SMAD2',
+    '4221':'MEN1',
+    '4763':'NF1',
+    '4771':'NF2',
+    '7157':'TP53', 
+    '5728':'PTEN', 
+    '5925':'Rb',
```
dhimmel: [RB1 is the approved symbol](https://www.ncbi.nlm.nih.gov/gene/5925).

dhimmel (commented):
Looks great to me. It does seem that a few of your gene names are non-standard. Can you switch to approved symbols? Then we'll be ready to merge!

explore/Classifier_results-different_genes.py
```diff
@@ -0,0 +1,291 @@
...
+ax_ts.legend((train_results_ts[0], test_results_ts[0], per_results_ts[0]), ('Train', 'Test','%'))
+
+autolabel(train_results_ts,ax_ts)
+autolabel(test_results_ts,ax_ts)
+autolabel(per_results_ts,ax_ts)
+
+
+# In[12]:
+
+# Plot for oncogenes
```
gwaybio, line 267: Fascinating... Generally, the performance is _worse_ for oncogenes - particularly MYC and SRC. MYC would be an interesting case to contrast how well we improve performance by incorporating copy number data (since MYC has copy number gains in several tumor types)

gwaybio: I suspect also that Cognoma will need to be smart about how CNVs are incorporated in the `Y` matrix. My thinking is that for oncogenes a copy number gain would define a 1 whereas for tumor suppressors a copy number loss would define a 1.

brankaj (commented):
I am still getting that deprecation warning even with the different placing of the ignore command :/

dhimmel:
> I am still getting that deprecation warning even with the different placing of the ignore command :/

Not sure about the warning. You should only have to run the `warnings.ignore` command once, before the code that causes the warning ever gets executed. Have you updated your sklearn to 0.18 (recently done in https://github.com/cognoma/machine-learning/pull/53). This may remove the error.

I'll merge this tomorrow morning, so if you fix the error by then cool, but if not no big deal.

dhimmel (approved):

brankaj:
Thank you guys for all the comments. I have updated sklearn before running the code, but the warning is still there. I will try to figure out what can be done regarding this issue.
I am also curious about the bad performance for oncogenes. I do not believe it is the classifier 'fault', but I will try different parameter values just to rule out this possibility. Other possibility would be that features or labels for genes like MYC are not that reliable. If this is the case, it is less likely that any classifier can help.

gwaybio:
> I do not believe it is the classifier 'fault', but I will try different parameter values just to rule out this possibility. Other possibility would be that features or labels for genes like MYC are not that reliable. If this is the case, it is less likely that any classifier can help.

These are my thoughts at the moment as well. I think we will need to more intelligently define our Y matrix - this is where I think we'll get the most performance gains. This is an active research question right now!

Issue #61, opened 2016-10-19
Add stronger support for pip


jessept:
cognoma runs into a specific import issue when installing directly from github:

```
>>> from cognoml import analysis
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Projects/miniconda3/envs/test_cognoma/lib/python3.5/site-packages/cognoml/analysis.py", line 11, in <module>
    from cognoml.classifiers.logistic_regression import grid_search
ImportError: No module named 'cognoml.classifiers'
```

This does not appear when installing in conda, but makes installing with pip more time-consuming. This would require some exploration into the files necessary to be able to pip install directly from github.

dhimmel:
@jessept can you post the pip command you used?

Also I edited your post to put the error message in a code block using triple backticks.

jessept:
@dhimmel Pip command used: 
`pip install git+git://github.com/cognoma/machine-learning`

Issue #57, opened 2016-10-12
Fix cognoml package


awm33:
Trying to run the cognoml package, I found a couple issues.
- The requirements are not in setup.py, so a consumer has to manually install them. I manually installed them by looking at the repo's environment.yml file. The following were needed before the next issue was raised.
  - pandas==0.18.1
  - numpy==1.11.1
  - scikit-learn==0.18.0
  - scipy==0.17.1
  
  Take a look at [https://packaging.python.org/requirements/](https://packaging.python.org/requirements/)
- Looks like cogoml.classifiers is not included in the package

```
>>> from cognoml analysis
  File "<stdin>", line 1
    from cognoml analysis
                        ^
SyntaxError: invalid syntax
>>> from cognoml import analysis
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/amadonna/Documents/ml-workers/env/lib/python3.5/site-packages/cognoml/analysis.py", line 11, in <module>
    from cognoml.classifiers.logistic_regression import grid_search
ImportError: No module named 'cognoml.classifiers'
```

Might have to be included [here](https://github.com/cognoma/machine-learning/blob/master/setup.py#L48).

I'd suggest testing the module in a clean environment.

Installed using `pip install git+git://github.com/cognoma/machine-learning`

awm33:
@dhimmel FYI

dhimmel:
@awm33 currently we only support installation into the conda environment specified in `environment.yml`. You may be able to get conda as simply as `pip install conda`. Note that conda may supersede your default Python path, but this can be undone.

I will look into also supporting installation via pip when requirements are not yet installed.

Regarding `ImportError: No module named 'cognoml.classifiers'` -- not sure / surprises me. I have never included submodules in the `packages` list in `setup.py`. I am pretty sure that the `__init__.py` in the directory is sufficient. Let me know if this issue persists. Could it be that the package isn't actually installed?

jessept:
@dhimmel @awm33 I just did a fresh install 5 minutes ago and was able to replicate the issue w the imports.

The issue is more about providing support for pip v conda and how the virtualenv implementations can differ. There is a note at the bottom of the readme here: https://github.com/cognoma/machine-learning

You need to download, then navigate to the project's root directory and do the pip install there rather than straight from the github site. This will put the project's root directory in your python path and be able to find the package correctly.

[...]

Issue #58, opened 2016-10-12
cognoml tests


awm33:
cognoml should really have tests. At the very least the "public" functions. The complex operations could be mocked.

dhimmel:
@awm33, I agree and will work on this.

In the past, I've found [pytest](http://doc.pytest.org/en/latest/) to be the most pythonic and user-friendly testing library. Anyone agree or disagree?

PR #62, opened 2016-10-20
Improve pip installation

dhimmel:
Specifies dependencies in `setup.py` so pip installation works outside of the conda `cognoma-machine-learning` environment. Closes https://github.com/cognoma/machine-learning/issues/57.

Closes https://github.com/cognoma/machine-learning/issues/61 and https://github.com/cognoma/machine-learning/issues/57 by resolving `ImportError: No module named 'cognoml.classifiers'`.


dhimmel:
## Evaluation

I evaluated whether pip installation works by:

Starting a python docker

``` sh
docker run --interactive --tty --entrypoint=/bin/bash --volume `pwd`/docker-downloads:/download python:3.5.2
```

Then install `cognoml` corresponding to this pull request using:

``` sh
pip install https://github.com/dhimmel/machine-learning/archive/pip.zip
```

In container shell, run `python`.

Then run the following python:

``` python
import json
import requests
from cognoml.analysis import classify

# Create a classifier using mock input. Print output to stdout.
url = 'https://github.com/cognoma/machine-learning/raw/876b8131bab46878cb49ae7243e459ec0acd2b47/data/api/hippo-input.json'
response = requests.get(url)
payload = response.json()
payload['data_version'] = 4
results = classify(**payload, json_sanitize=True)
json_results = json.dumps(results, indent=2)
```

Getting the following error (which has been [experienced by others](https://www.kaggle.com/forums/f/15/kaggle-forum/t/22023/no-space-left-on-device-running-sklearn-in-docker-python-container/125987)):

```
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/numpy_pickle.py", line 92, in write_array
    pickler.file_handle.write(chunk.tostring('C'))
OSError: [Errno 28] No space left on device
```

Which I think results from joblib writing to disk and exhausting the container's disk quota. Attempted to increase disk quota by adding a `--storage-opt size=120G` argument to docker run, but got the following error:

```
docker: Error response from daemon: --storage-opt is not supported for aufs.
```

Looks like I may have to [switch my docker storage backend](http://stackoverflow.com/a/39891717/4651668).

dhimmel:
I changed my docker storage device using:

``` sh
service docker stop
sudo dockerd --storage-driver=devicemapper
```

Then I was able to set `--storage-opt` using:

```
docker run --interactive --tty --entrypoint=/bin/bash --volume=`pwd`/docker-downloads:/download --storage-opt size=1000G python:3.5.2
```

However, I'm still running into the error. Here is the full trace:

```
>>> results = classify(**payload, json_sanitize=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/cognoml/analysis.py", line 64, in classify
    grid_search.fit(X=X_train, y=y_train)
  File "/usr/local/lib/python3.5/site-packages/sklearn/model_selection/_search.py", line 940, in fit
    return self._fit(X, y, groups, ParameterGrid(self.param_grid))
  File "/usr/local/lib/python3.5/site-packages/sklearn/model_selection/_search.py", line 562, in _fit
    for parameters in parameter_iterable
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/parallel.py", line 768, in __call__
    self.retrieve()
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/parallel.py", line 719, in retrieve
    raise exception
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/parallel.py", line 682, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
  File "/usr/local/lib/python3.5/multiprocessing/pool.py", line 608, in get
    raise self._value
  File "/usr/local/lib/python3.5/multiprocessing/pool.py", line 385, in _handle_tasks
    put(task)
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/pool.py", line 371, in send
    CustomizablePickler(buffer, self._reducers).dump(obj)
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/pool.py", line 240, in __call__
    for dumped_filename in dump(a, filename):
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/numpy_pickle.py", line 483, in dump
    NumpyPickler(f, protocol=protocol).dump(value)
  File "/usr/local/lib/python3.5/pickle.py", line 408, in dump
    self.save(obj)
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/numpy_pickle.py", line 277, in save
    wrapper.write_array(obj, self)
  File "/usr/local/lib/python3.5/site-packages/sklearn/externals/joblib/numpy_pickle.py", line 92, in write_array
    pickler.file_handle.write(chunk.tostring('C'))
OSError: [Errno 28] No space left on device
```

It would suck if we couldn't use sklearn's parallel processing when using Docker. But let's deal with this elsewhere.

dhimmel:
@awm33 and @jessept, can you review this pull request? Are you able to pip install in a virtualenv and run the python code above?

Also currently `setup.py` specifies:

``` python
install_requires=[
    'numexpr',
    'numpy',
    'pandas',
    'requests',
    'scikit-learn >= 0.18.0',
    'scipy',
]
```

Should we lock down pakcage versions more with the downside of incompatibility with most python installations? In other words, should users have to install this package in a designated environment to guarantee identical versions.

awm33:
@dhimmel joblib's default directory is /tmp/joblib, you need to tell it to use the volume you mounted. You can do this in code if you are using joblib directly, but we are using it through scikit, which doesn't let you pass that through to the joblib.Parallel instance it uses. But luckily joblib lets you set an env variable called JOBLIB_TEMP_FOLDER.

I was able to run the cognoml module using

Dockerfile

```
FROM python:3.5.1
ENV PYTHONUNBUFFERED 1
RUN pip install https://github.com/dhimmel/machine-learning/archive/pip.zip
```

Then run

``` sh
docker build --tag cognoml-test .
docker create --interactive --tty --volume /data --name test1 cognoml-test
docker start test1
Andrews-MacBook-Pro:ml-workers amadonna$ docker exec --interactive --tty test1 bash
root@6c4ef8cdf09c:/# JOBLIB_TEMP_FOLDER=/data/joblib python
Python 3.5.1 (default, Jun  9 2016, 15:37:54) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> import requests
>>> from cognoml.analysis import classify
>>> url = 'https://github.com/cognoma/machine-learning/raw/876b8131bab46878cb49ae7243e459ec0acd2b47/data/api/hippo-input.json'
>>> response = requests.get(url)
>>> payload = response.json()
>>> payload['data_version'] = 4
>>> results = classify(**payload, json_sanitize=True)
Downloading https://ndownloader.figshare.com/files/5864859 to `expression-matrix.tsv.bz2`
Downloading https://ndownloader.figshare.com/files/5864862 to `mutation-matrix.tsv.bz2`
Downloading https://ndownloader.figshare.com/files/5864865 to `samples.tsv`
>>> json_results = json.dumps(results, indent=2)
```

awm33:
@dhimmel Generally it's a good idea to lock down dependency versions. 

>  downside of incompatibility with most python installations

I'm not sure what you mean. Without the version numbers, it will just download the latest, it also looks like it works with the latest, but that could change.

dhimmel:
@awm33 elegant solution for the storage issue! Do you disable buffering in Python (`ENV PYTHONUNBUFFERED 1`) since it's usually more annoying than helpful or is this specifically needed in our case?

> I'm not sure what you mean. Without the version numbers, it will just download the latest, it also looks like it works with the latest, but that could change.

I agree downloading the latest will eventually lead to things breaking. However, I see some users wanting to use the cognoml package as part of their analysis pipelines. If we lock down the versions so specifically, it essentially precludes it's use outside of a dedicated environment?

In any case, I think we should start out with the safe version (specifying exact versions) and then consider expanding in the future when we have tests and continuous integration in place (#58).

awm33:
@dhimmel I was setting `PYTHONUNBUFFERED 1` because some Docker documentation I was reading used it. I tried removing it, but then the interactive console stopped displaying output.

I think the lock down sounds reasonable. A user can also pass `--no-deps` to pip to ignore the dependencies and install them manually if they would like other versions. You can add that to the README if you don't think people will realize that.

One request, that can be in another PR if you'd like. Can the `data_directory` be an optional kwarg to classify? Then we can create a data volume and place everything there. `downloads` can be kept as the default.

dhimmel:
Verifying that the example code fully executed in the Docker on e7d11976b1ec4b7975bd2d0777b1636ec9068e93. @awm33 does this PR look ready to you?

awm33:
@dhimmel 👍

jessept:
@awm33 I have started a refactor of the data download and classify processes that includes the 'download' option you are talking about. I haven't submitted a pull request as I haven't finished yet but you can view the forked code on my github. I also added a few performance improvements and started moving common operations to classes instead of all functions.

dhimmel:
@jessept good to hear. Note that we've moved cognoml to it's own repo: [`cognoma/cognoml`](https://github.com/cognoma/cognoml), see https://github.com/cognoma/machine-learning/issues/64. It would be best if you could open your pull request there. Also note that it's okay to open a pull request when the feature is still in progress. Just add `WIP` to the pull request title.

Also @jessept and @stephenshank, feel free to review open pull requests on `cognoma/cognoml` -- we need to get a maintainance up and running there.

PR #63, opened 2016-10-23
Cross validation evaluation

gwaybio:
Here, I add functionality to output the cross validation performance of the best classifier. The metric is a useful comparison across models and disease types - especially b/c of imbalanced class sizes and high heterogeneity in models.


scripts/3.TCGA-MLexample_Pathway.py
```diff
@@ -87,39 +87,45 @@
 
 # In[7]:
 
+# Perform 5 fold cross validation
+folds = 5
```
dhimmel, line 91: Noting that previously 3 folds were used (default). This changes the number of folds to 5, which will result in better performance at the cost of additional computation.

scripts/3.TCGA-MLexample_Pathway.py
```diff
@@ -313,11 +319,21 @@ def get_threshold_metrics(y_true, y_pred, tissue='all'):
 metrics_test = get_threshold_metrics(y_test, y_pred_test)
 
 
-# In[33]:
+# In[34]:
+
+# Rerun "cross validation" for the best hyperparameter set to define
+# cross-validation disease-specific performance. Each sample prediction is
+# based on the fold that the sample was in the testing partition
+y_cv = cross_val_predict(cv_pipeline.best_estimator_, X=X_train, y=y_train,
```
dhimmel, line 327: Didn't know this was standard practice, as you are grouping predictions that were made from different models. However, given that sklearn has a `cross_val_predict` function, I'll accept its validity. The alternative would be to draw an ROC curve for each CV-fold, but this is much prettier and concise. It appears to work here.

dhimmel: I'm worried that this computation has already been done. Can we somehow extract the cross-validated predictions from `cv_pipeline` rather than recomputing them?

dhimmel: Also this may be worth adding to `cognoml`, but only if we don't have to refit.

scripts/3.TCGA-MLexample_Pathway.py
```diff
@@ -341,22 +357,23 @@ def get_threshold_metrics(y_true, y_pred, tissue='all'):
     y_tissue_pred_train = y_pred_train[y_train.index.isin(sample_sub)]
```
dhimmel, line 357: Are all of these index selections redundant? Could we use a single boolean indexer?

dhimmel: If you created a dataframe with each observation and its training, cv, and testing predictions as well as its disease, this would be much cleaner?

The principle being never reorder or subset data separately when it can be performed jointly.

dhimmel (changes requested):
Nice, I think you pick up on an additional performance curve that we should consider widely adopting (the CV ROC curve).

rdvelazquez:
I was wondering if there was a feature like this but had never seen `cross_val_predict`. I think this really helps address the issue of the testing AUROC being highly dependent on the chosen random state, especially for data sets with few mutations. Great suggestion @gwaygenomics!

> I'm worried that this computation has already been done. Can we somehow extract the cross-validated predictions from cv_pipeline rather than recomputing them?

I did a little bit of digging to see if there is a way to do this in scikit-learn and didn't find anything. There was an issue for this https://github.com/scikit-learn/scikit-learn/issues/5030 but it was closed in 2016. Considering that (in this instance) cv fit and predict are already performed 16 times (4 alphas x 4 L1_ratios), I don't think doing it one more time is too big of a performance hit.

`cross_val_predict` seems like a good addition to this notebook and I'd be in favor of adding it to `cognoml` as well. Let me know how I can help.

dhimmel:
> Let me know how I can help.

@rdvelazquez, will do. Currently traveling so probably won't get a chance until next Tuesday or Wednesday.

rdvelazquez:
@gwaygenomics I liked the idea for this performance curve as well but I think this won't get included in cognoma right now. Are you ok if I close this PR?

gwaybio:
:+1:

Issue #56, opened 2016-10-11
Selecting the elastic net mixing parameter


dhimmel:
Thus far we've been using grid search (cross validation) to select the optimal elastic net mixing parameter. For `SGDClassifier`, this mixing parameter is set using `l1_ratio`, where `l1_ratio = 0` performs ridge regularization and `l1_ratio = 1` performs lasso regularization.

Here's what I'm thinking:

Grid search is not the appropriate way to select the mixing parameter. Ridge (with the optimal regularization penalty, `alpha`) will always perform better than the optimal Lasso. The reason is that there's a cost for the convenience of sparsity. Lasso makes difficult decisions about which features to select. Therefore the sparsity can aid in model interpretation, but weakens performance because identifying only the predictive features is an impossible task.

For example, see our grid from [this notebook](https://github.com/cognoma/machine-learning/blob/5199430cf82bcb3763084e559e4f6d563df656a5/algorithms/SGDClassifier-master.ipynb) (note this used MAD feature selection to select only 500 features which likely accentuates the performance deficit as `l1_ratio` increases).

![grid](https://cloud.githubusercontent.com/assets/1117703/19273374/4ec6f042-8f9a-11e6-95f9-51963693fb2b.png)

So my sense is that `l1_ratio` should be chosen based on what properties we want the model to have, not based on maximum CV performance. If we only care about performance, we might as well save ourselves the computation time and always go with ridge or the default `l1_ratio = 0.15`. `l1_ratio = 0.15` can still filter ~50% of features with little performance degradation. But if you want real sparsity (lasso) there's going to be a performance cost -- and the user not grid search will have to make this decision.

dhimmel:
Also, I'd rather spend more time optimizing `alpha` (regularization strength). [glmnet in R](https://web.stanford.edu/~hastie/glmnet/glmnet_alpha.html) defaults to trying a sequence of 100 different regularization strengths.

cgreene:
Agree with @dhimmel about ridge/lasso trade-offs. We could ask the user how much they value sparsity vs performance if we can figure out a way that's not too confusing.

gwaybio:
> So my sense is that l1_ratio should be chosen based on what properties we want the model to have, not based on maximum CV performance.

Agreed!

patrick-miller:
If we are performing PCA on the expression matrix to create our features, then I am not sure how important sparsity is going to be in the end classifier. This is probably even more true when the number of components we chose is <= 100.

[...]

PR #53, opened 2016-09-30
Upgrades: sklearn to 0.18.0, data to v5

dhimmel:
Update data to latest available on figshare (v5). Expression and mutation have not changed since v4, so do not expect changes from this upgrade.

Update sklearn to 0.18.0. See [what's new](http://scikit-learn.org/0.18/whats_new.html). Particularly exciting is:

> Fix incomplete `predict_proba` method delegation from `model_selection.GridSearchCV` to `linear_model.SGDClassifier` (scikit-learn/scikit-learn#7159) by Yichuan Liu (@yl565).

`3.TCGA-MLexample_Pathway.ipynb` had incorrect data location. Fixed so all root notebooks would execute.

PR #67, opened 2016-11-01
Marginal gain of gene expression data over covariates

joshlevy89:
Compares a covariates only model to a covariates with gene expression model in predicting TP53 mutations.


dhimmel:
AWESOME! Just two things before I review.

Can you remove the space in `explore/marginal-gain/1. Marginal-Gain-TP53-Example.ipynb`?

Then run

```
cd explore/marginal-gain
jupyter nbconvert --to script 1.Marginal-Gain-TP53-Example.ipynb
```

and track the exported `.py` file. We use the py file for GitHub diffing and line comments.

joshlevy89:
I added a notebook (and script) that extends the comparison of covariates only to covariates+expression models to multiple mutations. If you look at the bar plot at the bottom, you'll see that the training auroc is always better for the covariates+expression model (positive values) but that the testing auroc drops off dramatically and is lower in every case other than TP53.  This is probably a case of overtraining.  Thoughts?

explore/marginal-gain/2.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,185 @@
...
+            rows.append(row)
+    df = pd.DataFrame(rows)
+    return df
+
+auroc_df = {}
+
+
+# In[46]:
+
+get_ipython().run_cell_magic('time', '', '# Train model a: covariates only.\nwarnings.filterwarnings("ignore") # ignore deprecation warning for grid_scores_\nrows = list()\nfor m in list(mutations):\n    series = pd.Series()\n    series[\'mutation\'] = m\n    series[\'symbol\'] = mutations[m]\n    rows.append(get_aurocs(X[\'a\'], Y[m], cv_pipeline[\'a\'], series))\nauroc_df[\'a\'] = pd.DataFrame(rows)\nauroc_df[\'a\'].sort_values([\'symbol\', \'testing_auroc\'], ascending=[True, False], inplace=True)')
```
dhimmel: How about more descriptive keys for `auroc_df` than `a`, `b`, and `c`? Also maybe rename `auroc_df` to `auroc_dfs` to make clear that it itself is not a dataframe.

explore/marginal-gain/2.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,185 @@
...
+auroc_df['c'][['mutation', 'symbol']] = auroc_df['b'].loc[:, ['mutation', 'symbol']]
+auroc_df['c']
+
+
+# # Covariates only vs covariates+expression model
+
+# In[68]:
+
+plot_df = pd.melt(auroc_df['c'], id_vars='symbol', value_vars=['mean_cv_auroc', 'training_auroc', 'testing_auroc'], var_name='kind', value_name='auroc')
+grid = sns.factorplot(y='symbol', x='auroc', hue='kind', data=plot_df, kind="bar")
```
dhimmel: Can you change the x-axis label from `auroc` to `Δ AUROC`, so it's clear that it's the change in AUROC?

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
```diff
@@ -0,0 +1,327 @@
...
+    ('imputer', Imputer()),    
+    ('select', SelectKBest(fs_mad)),
+    ('standardize', StandardScaler()),
+    ('classify', SGDClassifier(random_state=0, class_weight='balanced'))
+])
+
+cv_pipeline = {}
+for k in ['a','b']:
+    if k == 'a': param_grid['select__k'] = ['all']
+    elif k=='b': param_grid['select__k'] = [2000]
```
dhimmel: I think you're likely selecting out most of the covariates because they have a low median absolute deviation. This is probably why the "covariate + expression" models are worse than the "covaraite only" models.

dhimmel: [`FeatureUnion`](http://scikit-learn.org/stable/modules/generated/sklearn.pipeline.FeatureUnion.html) may be the correct way to deal with this.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
```diff
@@ -0,0 +1,327 @@
...
+# ## Define pipeline and Cross validation model fitting
+
+# In[22]:
+
+# Parameter Sweep for Hyperparameters
+param_grid = {
+    'classify__loss': ['log'],
+    'classify__penalty': ['elasticnet'],
+    'classify__alpha': [10 ** x for x in range(-3, 1)],
+    'classify__l1_ratio': [0, 0.2, 0.8, 1],
```
dhimmel: Let's stick to a single `l1_ratio` but try more `alpha` values (see https://github.com/cognoma/machine-learning/issues/56):

``` python
    'classify__alpha': 10.0 ** np.linspace(-3, 1, 10),
    'classify__l1_ratio': [0.15],
```

The above values are from [`logistic_regression.py#L19-L20`](https://github.com/cognoma/cognoml/blob/f9669edfe0871cd161ffefa3409f67c98dd407a2/cognoml/classifiers/logistic_regression.py#L19-L20).

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
```diff
@@ -0,0 +1,327 @@
...
+            (coef_df[k].weight == 0).mean(),
+            (coef_df[k].weight < 0).sum(),
+            (coef_df[k].weight > 0).sum()
+    ))
+
+
+# In[263]:
+
+# What are the top weighted features for model a and model b?
+display(coef_df['b'].head(5))
```
dhimmel: Can we add `n_positives` and `n_negatives` to these dataframes? Would help me diagnose the variability in performance between CV, training, & testing.

dhimmel (changes requested):
Cool, I think there are some weird things going on that make these results hard to interpret. Hopefully, addressing my comments will clear things up!

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
joshlevy89: Sounds good.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
joshlevy89: Ah, good catch. 

I implemented it two ways. 1) Using FeatureUnion.  This seems to work but is a bit clunky. 2) Using DataFrameMapper. Cleaner syntax and I think it works, but getting a warning. Let me know which approach you prefer.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
joshlevy89: For which dataframe? coef_df? I think I'm misunderstanding b/c seems like that wouldn't help--the weights already tell you that. Maybe you mean for auroc_df or something?

explore/marginal-gain/2.Marginal-Gain-Multiple-Mutations.py
joshlevy89: sure.

explore/marginal-gain/2.Marginal-Gain-Multiple-Mutations.py
joshlevy89: yes. a and b are not terribly descriptive but they actually stand for models a and b, which I have consistently used. I changed c though b/c that is totally not description to diff_ba (difference between model b and a)

joshlevy89 (commented):
I have modified the code but still need to run it. Once I have done so, I will repost with changes.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
dhimmel: My mistake... yeah I'm referring to `auroc_df` in the second notebook.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
dhimmel: Cool... What's the warning for `DataFrameMapper` -- I didn't see a warning in your notebook. I agree `DataFrameMapper` is cleaner. Do you think the added cleanliness is worth the extra dependency?

Can you modify `2.Marginal-Gain-Multiple-Mutations.ipynb` to use your preferred method?

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
```diff
@@ -0,0 +1,327 @@
...
+
+# ## Define pipeline and Cross validation model fitting
+
+# In[22]:
+
+# Parameter Sweep for Hyperparameters
+param_grid = {
+    'classify__loss': ['log'],
+    'classify__penalty': ['elasticnet'],
+    'classify__alpha': [10 ** x for x in range(-3, 1)],
```
dhimmel: Looks like your notebook only has 1 value here. Reminder to reset when you're ready for a final analysis. I think it's also fine to consolidate to a single notebook later (probably the current `2.`).

explore/marginal-gain/2.Marginal-Gain-Multiple-Mutations.py
dhimmel: Do what you think is best. My opinion is that it's almost always better to be explicit and save the reader from having to remember a "labeling dictionary" of sorts.

dhimmel (commented):
Awesome work, especially dealing with the difficulties of parrallel feature processing -- it looks like the sklearn design didn't fully anticipate this crucial use case.

This PR is close to ready... now time to apply your solutions to the second notebook.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
joshlevy89: Will do. Just for testing.

dhimmel:
Is it possible to only apply the Imputer to the covariate features? The gene expression doesn't have missing data, so it may speed things up.

explore/marginal-gain/1.Marginal-Gain-TP53-Example.py
joshlevy89: I will go with FeatureUnion. The error from DataFrameMapper (you did not see the warning because ran other method) was that the estimator "Estimator %s modifies parameters in **init**." from the sklearn base library (http://contrib.scikit-learn.org/imbalanced-learn/_modules/sklearn/base.html).  Could never figure that one out, but also the DataFrameMapper lib was finicky (e.g. the mapper needed to be at top of pipeline). Overall, I think it is not worth the extra dependency, as you suggested. I modified the FeatureUnion method to use lambda method in FunctionTrasformer which makes it a bit more concise).

explore/marginal-gain/2.Marginal-Gain-Multiple-Mutations.py
joshlevy89: Will change to model a and model b

joshlevy89:
It should be possible to apply Imputer to only covariates. But for some reason this is not working with FeatureUnion. It's as if FunctionTransformer were were working for SelectKBest but not Imputer. I will run the code through and then maybe you can have a look at restructuring the code to include this change? Could be I'm missing something.

joshlevy89:
I ran the code on a restricted dataset (1000 samples) for all the mutations. The results are in the 1.Marginal-Gain-Multiple-Mutations notebook. As you can see, the testing cv's are still lower for the combined model for a number of mutations. 

To diagnose, I added n_positives and n_negatives as you asked. I also added some statistics on the ranks of the weights for covariates wrt ranks of all features.  Cumulative, median and mean ranks are presented--cumulative rank means the total index positions of covariate features in the sorted coef_df dataframe. 

I can run this on the whole dataset, but I think I will hit the same issue. 

(Also, note that to make this run, I have set the 1st value of the y_test and y_train to 1 just to ensure at least one positive mutation example. Not sure--could be affecting the results a bit.).

joshlevy89:
After writing that, I realized that I did not change the alpha parameter back to test all values. Let me do that and get back to you. Probably a regularization issue.

joshlevy89:
OK, I re-ran it with all hyperparamaters and data.  Still, the testing is only substantially better for TP53. Strangely, the training auroc is lower for some mutations--I don't even know how that's possible. Hopefully the updates will make it easier to diagnose.

1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,257 @@
...
+
+# In[18]:
+
+display(auroc_dfs['model a'])
+display(auroc_dfs['model b'])
+
+
+# In[14]:
+
+auroc_dfs['model a'].to_pickle('./covariates_only.pkl')
```
dhimmel: Any reason for the switch to pickles over TSV here -- TSV is preferable so its readable outside of Python. 

Consider combining both tables into a single dataframe for storage by adding a `model` column and then concatenating.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,253 @@
...
+    y_pred_train = pipeline.decision_function(X_train)
+    y_pred_test = pipeline.decision_function(X_test)
+    cv_score_df = grid_scores_to_df(pipeline.grid_scores_)
+    coeff_df = get_coeffs(pipeline, X_train, model_type)
+    n_pos, n_neg = get_sign_coeffs(coeff_df)
+    cov_ranks = get_ranks_covariates_feat(coeff_df)
+    series['mean_cv_auroc'] = cv_score_df.score.max()
+    series['training_auroc'] = roc_auc_score(y_train, y_pred_train)
+    series['testing_auroc'] = roc_auc_score(y_test, y_pred_test)
+    series['n_pos_coeffs'] = n_pos
```
dhimmel: The new columns `n_pos_coeffs` and `n_neg_coeffs` are really useful. They're not actually what I meant by `n_positives` and `n_negatives` -- but let's definitely keep them.

By `n_positives` I meant number of samples with the mutation (`sum(y == 1)`).

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,253 @@
...
+# In[83]:
+
+# Get statistics for a given model. 
+
+def get_aurocs(X, y, pipeline, series, model_type):
+    """
+    Fit the classifier for the given mutation (y) and output predictions for it
+    """
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
+    y_train[0] = 1
```
dhimmel: > (Also, note that to make this run, I have set the 1st value of the y_test and y_train to 1 just to ensure at least one positive mutation example. Not sure--could be affecting the results a bit.).

@joshlevy89, I'm concerned about this workaround. I think specifying the `stratify=y` in [`train_test_split`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html) should resolve this problem. If not, I don't think we want to make a workaround -- the failure shouldn't be masked.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,253 @@
...
+
+# In[83]:
+
+# Get statistics for a given model. 
+
+def get_aurocs(X, y, pipeline, series, model_type):
+    """
+    Fit the classifier for the given mutation (y) and output predictions for it
+    """
+    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
```
dhimmel: Let's increase testing to 25% (`test_size=0.25`). Our testing AUROCs are potentially super unstable due to the small number of positives.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,253 @@
...
+        ('covariates_feats', covariate_feats),           
+        ('expression_feats', expression_feats)
+    ])),
+    ('classify', SGDClassifier(random_state=0, class_weight='balanced'))
+])
+
+covariates_pipeline = Pipeline(steps=[
+    ('imputer', Imputer()),
+    ('standardize', StandardScaler()),
+    ('select', SelectKBest(fs_mad,'all')),
```
dhimmel: Let's just delete this entire step if its not actually doing any selection.

dhimmel (commented):
I made a few additional comments. Additionally, it looks like you may have run a `git add --all` that added some files we don't want to track.

Specifically some `.DS_Store` files crept in which should be removed.

Also `1.Marginal-Gain-Multiple-Mutations.py` appears to have been uploaded in two directories.

1.Marginal-Gain-Multiple-Mutations.py
joshlevy89: Done.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
joshlevy89: I assume you just mean the ('select', SelectKBest(fs_mad,'all')) step. I will take it out (and modify how the coef_df dataframe is built to handle lack of 'select' step).

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
joshlevy89: Yes.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
joshlevy89: Don't worry, that was just a workaround during debugging--I took it out when I ran the script (I think you are looking at an old notebook here.) In any case, I like your method better.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
joshlevy89: Gotcha. Updated. As you might expect, the combined model shows the clearest advantage where we have a lot of positive examples to work from (TP53).

joshlevy89:
I've made all the requested changes. I will run overnight and see how the results turn out.

joshlevy89:
...and I will upload after that.

dhimmel:
> I will run overnight and see how the results turn out.

Awesome! And when you are ready for a final review, just remove "WIP" from the issue title.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.ipynb
```diff
@@ -0,0 +1,1299 @@
...
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "auroc_dfs['model a']['model'] = 'covariates_only'\n",
+    "auroc_dfs['model b']['model'] = 'combined'\n",
+    "auroc_df = pd.concat([auroc_dfs['model a'],auroc_dfs['model b']])\n",
+    "display(auroc_df)\n",
+    "auroc_df.to_csv(\"./auroc_df.tsv\", sep=\"\\t\")"
```
dhimmel: If not too late, can you add `float_format='%.3g'` in `to_csv` to improve readability. Also specify `index=False`.

joshlevy89: Done. I uploaded the updated notebook and script. Ignore the output (have not rerun) but let me know if there is anything else you'd like me to change before I run tonight.

1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,257 @@
+
```
dhimmel: Wrong location again -- we really should explore a CI server to automatically make these exports.

```
git mv 1.Marginal-Gain-Multiple-Mutations.py  explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```

joshlevy89: Ugh, right. Keep forgetting to cd.

joshlevy89:
OK, ran data through most recent config. I think some of the changes helped. Not exactly sure what to make of some results though. Like why the divergence between "mean_cv_auroc" and training_auroc. Why is training/testing auroc .5 for ALK and BRAC1. 

Also, I borrowed some of that code...so I'm not as clear on it. It seems like the calculation of mean_cv_auroc as "cv_score_df.score.max()" may not be what we want. Again, didn't write that section so not as sure.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,259 @@
...
+    series['n_pos_coeffs'] = n_pos
+    series['n_neg_coeffs'] = n_neg
+    series['n_positive_mutation'] = sum(y==1)
+    series['n_negative_mutation'] = sum(y==0)
+    series['cum_rank_cov_feat'] = cov_ranks.sum()
+    series['median_rank_cov_feat'] = np.median(cov_ranks)
+    series['mean_rank_cov_feat'] = np.mean(cov_ranks)
+    return series
+
+def grid_scores_to_df(grid_scores):
```
dhimmel: Since you're using sklearn 0.18, there's a slightly simpler way of extracting a CV dataframe. See [here](https://github.com/cognoma/machine-learning/pull/53/files#diff-2e04242337bdd5fc54de4c2c8ccd8879R149). If you switch to the new method, you can take the max of the `mean_test_score` column to get `mean_cv_auroc`.

explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.py
```diff
@@ -0,0 +1,259 @@
...
+    return ranks
+
+auroc_dfs = {}
+
+
+# ## Train the models.
+
+# In[12]:
+
+get_ipython().run_cell_magic('time', '', '# Train model a: covariates only.\nwarnings.filterwarnings("ignore") # ignore deprecation warning for grid_scores_\nrows = list()\nfor m in list(mutations):\n    series = pd.Series()\n    series[\'mutation\'] = m\n    series[\'symbol\'] = mutations[m]\n    rows.append(get_aurocs(X[\'model a\'], Y[m], cv_pipeline[\'model a\'], series, \'model a\'))\nauroc_dfs[\'model a\'] = pd.DataFrame(rows)\nauroc_dfs[\'model a\'].sort_values([\'symbol\', \'testing_auroc\'], ascending=[True, False], inplace=True)\ndisplay(auroc_dfs[\'model a\'])')
```
dhimmel: Let's remove the `%%time` magic. While very helpful, I think line by line commenting on the diff here will be more usefull.

dhimmel:
@joshlevy89 thanks for grinding forward on this PR. We're definitely experiencing some unexpected turbulence. Commenting on analysis at 57fff61d9ca1c0e475e74dcd16a3a9eb788bce16:

Something wacky is going on. In [`auroc_df.tsv`](https://github.com/joshlevy89/machine-learning/blob/57fff61d9ca1c0e475e74dcd16a3a9eb788bce16/explore/marginal-gain/auroc_df.tsv), it the ALK and BRCA1 models have no nonzero coefficients. In other words, the model makes the same prediction for every sample, hence an AUROC of 0.5. However, in past covaraite only models, we've been able to fit models for ALK and BRCA1 -- see https://github.com/cognoma/machine-learning/pull/47#issuecomment-248044907. ALK has 138 positives while BRCA1 has 136, so I don't think this is primarily a sample size issue. Interestingly, the `mean_cv_auroc` performance for ALK and BRCA1 is in line with the past covaraite only models. Therefore I think there is an issue that occurs after cross validation that is messing up the final estimator.

I think one approach is to create a much simplified analysis and then take it from there. I'd recommend the following analysis:
1. limit to the following covariates (removing need for imputation): `n_mutations_log1p` and anything starting with `acronym_` or `gender_`.
2. created combined X matrix in pandas before machine learning. Don't rely on the pipeline to do this since the pipeline may have erratic behavior. 
3. remove MAD feature selection. The pipeline will now only have standardization and classification steps.

I'm hoping this will get us to something that works, where we can then build upwards and know what exactly is breaking. Feel free to open a new PR if you want your current hardwork to stay at the tip of the branch for this PR. Or you can also make a new branch from 57fff61d9ca1c0e475e74dcd16a3a9eb788bce16 at any time to get back to here (to preserve the pipeline work).

I see current total runtime is high for the combined cell (Wall time: 1h 19min 27s). I suspect this makes development difficult. For dev, I'd recommend speeding things up by shortening the list of `mutations` dict to one or two.

joshlevy89:
I think you hit it on the head. I created a branch with simplified analysis and just ran the covariates on ALK. (Luckily, in some sense, the covariates has the .5 auroc problem too so it's fast to debug).

I get the same problem vastly simplifying the pipeline. Check out the branch here: https://github.com/joshlevy89/machine-learning/blob/simplified/explore/marginal-gain/1.Marginal-Gain-Multiple-Mutations.ipynb. As you suggest, the covariates model is outputting the positive decision values for each observation. So I think there is something rather trivial going on, but not quite sure what. As in I don't think this has to do with the complexities of pushing sklearn to its limits. Thoughts?

joshlevy89:
(I will continue to play around...I think I can localize the error.)

joshlevy89:
Aha! We created one problem when trying to debug another. Remember when we decided to restrict our analysis to one il_ratio parameter? Well, unfortunately the chosen value did not work at all. I used the parameters from the notebook #47 and got back very similar values for ALK. 

I don't really understand why this results in .5 auroc for training and testing. It seems like those scores should follow the cv auroc. So I know where the issues exists (and how to fix), but not why. Thoughts on that?

I think the next moves are:
1) Verify that the approach works on a small sample (ALK and TP53). Probably select a good il_ratio for covariates and then use that for the combined model (because otherwise training will take forever).
2) Expand to all mutations.

joshlevy89:
For completeness (and sorry for all the messages) I noticed the the issue was different for BRAC1. There, it was the stratify (another thing we added during debugging!) in the train_test_split that caused it to output .5 for train and test.

dhimmel:
Can you add a [`VarianceThreshold`](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.VarianceThreshold.html) to the pipeline? This should remove features where all the values are the same... I wonder if zero variance features are causing the model fitting to fail so miserably.

joshlevy89:
I tried variance threshold for the covariates features and it had no effect when stratify was being used. Basically, stratify caused BRAC1 to fail (.5 auroc for train/test) no matter what l1 params were being used and ALK fail for certain l1 params.  Variance threshold did not change that. 

Last night, I tried running the code without stratify. As I had seen, it worked fine for covariates. But it did not for the combined data. In the case of RET, it failed (i.e. gave .5/.5 for train/test auroc). 

One check would probably be to just use the exact same pipeline for covariates and the combined data (ie remove select K best, and feature union) and run RET. That way, if RET is still messed up for the combined data, we will know it has something to do with the data itself and not the difference in pipelines. I will try that tonight.

joshlevy89:
I have some simplified, interesting results. I created a branch called simplified and put 3 notebooks in it. They all involve comparison of covariates only to combined model for Ret mutation.  Note that the pipelines are identical in both cases to ensure that differences are not due to that. 

1.Ret-GridSearch.ipynb: contains a small grid search. You can see that the covariates and the combined models select a different alpha ratio (.1 for covariates, 1 for combined). Covariates works, but combined model outputs .5/.5 for train/test.

2.Ret-NoGridSearch-Alpha1.ipynb: I got rid of the grid search to see if that was causing the problem (we already didn't think so as the mean_cv_roc was always reasonable). I set alpha to 1--and, as expected, the combined model failed. Not a grid search problem, per se, therefore. 

3.Ret-NoGridSearch-Alpha.1.ipynb: Same as 2) but set alpha to .1. Now, both models output reasonable values.  Remember, in grid search in notebook one, the combined model selected alpha=1, not alpha=.1.  So interesting that alpha=.1 in this notebook actually works better than alpha=1 in notebook 2. 

This leads me to the conclusion that somehow the best parameters are not being selected after grid search.  Because if they were, alpha = .1 would be selected in both cases. 

As a side note, in notebook 3 the test/train auroc for the combined model is still worse than the covariates model. Strange. It is more data with the exact same pipeline. My only guess is that the large number of features creates a highly non-convex space and the optimizer gets stuck in a local minimum. 

Finally, I think my computer is not really strong enough to run these analyses easily. I have had freeze ups several times just from running even the vastly reduced model.  I think I need to upgrade my hardware, but in the meantime I'm not really sure I can efficiently continue on this particular project. 

@dhimmel Have you had a chance to think about this?

joshlevy89:
Steps to fix:

- l1 ratio to zero (ridge regression)
- dimensionality reduction
- more alpha values
- more viz of mean_cv_auroc

dhimmel:
@joshlevy89, crossing fingers that the steps in https://github.com/cognoma/machine-learning/pull/67#issuecomment-265325874 will resolve this! Can't wait to see.

joshlevy89:
@dhimmel I have updated this pull request with the requested changes, and added some other stuff as well.  We are no longer experiencing fitting failures/turbulence when working with the expression data, probably due to adding the feature selection/pca.  :smiley: 

A few notes on what I have done here:
* There are three notebooks: **marginal-gain-Ret-mad.ipynb, marginal-gain-multiple-genes-mad.ipynb, marginal-gain-multiple-genes-pca.ipynb**
* In all three notebooks, I run fitting and visualization for three data sets:
  * covariates only
  * expression only
  * combined (covariates + expression)
* In **marginal-gain-Ret-mad.ipynb**, I use these datasets to predict Ret mutations, as an example, with parameters used in Branka's PR #52.  
  * Doing this, I get identical training/testing aurocs for the expression only data, thus verifying that the pipeline is working as expected.  
  * Note that feature selection with fs_mad is used here.
  * Notably, the combined dataset does not substantially improve prediction over covariates only. 
* In **marginal-gain-multiple-genes-mad.ipynb**, I do this same thing for all mutations in Branka's PR #52, except that l1_ratio is set to 0 for expediency.
  * Note that this dataset is larger than what I was using before: it contains all the mutations found in #47, except for CD274, and many others as well. 
  * As before, the combined dataset does not substantially improve prediction over covariates only.
* In **marginal-gain-multiple-genes-pca.ipynb**, I drop fs_mad selection in favor of PCA (the prior notebooks should nevertheless be kept, in my opinion, because they offers a comparison to #52).
  * I chose to follow hctai's approach in #71 for picking components.  In that PR, he mentions that the optimal number of components ranges from 30-100, so I split the difference and chose 65.  Down the line, if we decide to incorporate his suggestion to add n_components as a hyperparameter, we can update this notebook as well.  
  * Using pca, there is a tiny (perhaps spurious) increase in performance for the combined data over the covariates only.  Perhaps improved pca methodology or hyperparameter tuning would improve that.
  * It seems to me that there is a relationship between the marginal gain in the mean_cv_auroc/testing_auroc (but not training_auroc) and the number of mutations for a given gene (shown at the bottom).

I think that this PR is now ready to be merged.

explore/marginal-gain/marginal-gain-multiple-genes-pca.py
```diff
@@ -0,0 +1,340 @@
...
+    '3791':'KDR', #receptor tyrosine kinases
+    '25':'ABL1', #Cytoplasmic tyrosine kinases
+    '6714':'SRC', #Cytoplasmic tyrosine kinases
+    '5894':'RAF1',#cytoplasmic serine kinases
+    '3265':'HRAS',#regulatory GTPases
+    '4609':'MYC',#Transcription factors
+    '2353':'FOS',#Transcription factors
+}
+
+mutations = {**genes_LungCancer, **genes_TumorSuppressors, **genes_Oncogenes}
```
dhimmel, line 145: Really cool way to combine dicts, didn't realize you could do this!

dhimmel:
Hey @joshlevy89 -- thanks for this huge PR!

Wanted to let you know it's on my radar, but I may not get to it for a little bit.

The next step will be understanding why expression does not improve upon the covariate models -- whether it's fundamentally not useful or because we need to change our modeling.

joshlevy89:
@dhimmel No problem. I agree that is the next step. Indeed, it seems like the current approaches do not improve upon covariates.

explore/marginal-gain/marginal-gain-multiple-genes-pca.py
```diff
@@ -0,0 +1,340 @@
...
+
+# In[42]:
+
+get_ipython().magic('matplotlib inline')
+plt.style.use('seaborn-notebook')
+
+
+# In[43]:
+
+get_ipython().run_cell_magic('time', '', "path = os.path.join('..', '..', 'download', 'covariates.tsv')\ncovariates = pd.read_table(path, index_col=0)")
```
dhimmel: Let's restrict to only the following covariates:

+ `n_mutations_log1p`
+ anything starting with `acronym_`

We've seen these are the most important. This way we can remove imputation, which won't currently work for categorical variables that have been dummied anyway.

dhimmel: We can remove `%%time` statements from every cell besides `.fit`, given how annoying it makes the diffing.

dhimmel: Can you rerun the download notebook, so we're working with v6 of the dataset?

dhimmel (commented):
Made some small suggestions to help us narrow in on the situation. Apply to all 3 analyses.

Also fine to proceed ahead with only 2 of 3 notebooks and delete the third -- the PCA one is most important I think. Up to you.

joshlevy89:
@dhimmel applied changes and re-ran. I deleted marginal-gain-Ret-mad.ipynb example notebook, as it was just to validate that everything was working correctly.  I only modified marginal-gain-multiple-genes-pca.ipynb as more changes may be on the way, and we may want to delete marginal-gain-multiple-genes-mad.ipynb later anyway.

dhimmel:
@joshlevy89 great work.

I think you've shown that the prediction of many mutations is not benefiting from expression data, once covariates are accounted for. However, some mutations still are such as AKT1, TP53, RB1, NF2, and HRAS. This is really good to know -- our approach may only apply to certain mutations.

I also think your last plots are really informative -- gene expression seems most likely to contribute when the number of positives is higher. So this leads me to believe that our model lacks the necessary power to detect gene expression signatures when the sample size is low, even when using PCA.

I think it makes sense to refocus our ML team to look more into fitting models with both covariates and expression. Thanks for completing this gigantic analysis with lots of back and forth!

Up to you whether you want to delete `marginal-gain-multiple-genes-mad.ipynb` or update it. After that I think it'll be time to merge.

cgreene:
One thing that I noticed - many of the genes where the expression seems to add information beyond the covariates seem to be cell cycle genes. The various Ras genes + NF1 where @gwaygenomics has worked a fair amount seem to be relatively high performing. P53, but this is not surprising as it is known to induce a big effect in the transcriptome. RB1, PTEN, AKT are sort of in the cell cycle hit list. This list *would* be enriched for cell cycle genes anyway as a cancer gene list, but these are the things that stuck out at me.

gwaybio:
i agree - this PR is monumental! nice work @joshlevy89 

It is a bit jarring to see how well covariates perform alone in many circumstances. Can you remind me what the covariates are? Dummy variable disease type and log(mutation count)? It may be good to show a `covariates.columns` if space permits in the jupyter notebook. 

Also, it could be good to eventually show what the coefficients are for some of the models - at the very least to double check that all the signal the gene expression adds is not washed out by giant covariate weights. It would also be interesting to see the weights of the disease type for each model. I suspect that for some gene mutations, especially those that are heavily skewed across cancer types, (e.g. NF1 - 11% GBM (17/149) but 0.6% in THCA (3/485)) the covariate model alone will just pick out disease types rather than individual samples. Another way to test this would be to build models using only a single disease type and then make the comparisons suggested in this PR.

Perhaps these suggestions belong in a separate PR? I do not want to stall merging!

dhimmel:
> Perhaps these suggestions belong in a separate PR?

Let's do a separate PR. I think @joshlevy89 was in the process of moving. However, @gwaygenomics raises some great questions. @joshlevy89 you are free to continue tackling them after we merge this PR. But let's save that for a future PR.

joshlevy89:
@dhimmel I applied the changes to the marginal-gain-multiple-genes-mad.ipynb notebook and re-ran.  It will be good to keep this as a reference, especially to Branka's PR #52 .  

Should be good to merge now.  Thanks for the help on getting this out!  I think it raises some interesting issues as pointed out by @cgreene and @gwaygenomics, and it will be interesting to pursue this further.

dhimmel (approved):

dhimmel:
Great work @joshlevy89! Will merge now.

PR #69, opened 2016-11-17
Passive Aggressive Classifier

KT12:
Followed template provided by @dhimmel to implement classification with multilayer perceptron.  All of the preprocessing steps remain the same.

The AUROC score worked for the CV violin plot and heat map but not for the ROC curve.  I have replaced the AUROC curve with an F1 Micro Score for all CV analysis.

I have also provided a manual log of best parameters.  This can be found in the section following model fitting.

Thank you for reviewing.


dhimmel:
Nice! Can't wait to take a look.

First, can you rename `WIP-MultilayerPerceptron-KT12 (F1micro).ipynb`, so it doesn't have any spaces or parentheses. We like to keep our file names shell friendly.

Also export a script file [as described here](https://github.com/cognoma/machine-learning/blob/7406e51f087932062a076dfbd5573c0d0fd8cbc2/algorithms/README.md).

dhimmel:
Make sure to also remove the now outdated `algorithms/WIP-MultilayerPerceptron-KT12 (F1micro).ipynb`.

KT12:
Renamed the file, deleted the old, and added the converted it to a script for diff viewing.

Sorry I missed the part about converting to a script.  Thank you!

algorithms/scripts/WIP-passiveaggressive-KT12.py
```diff
@@ -0,0 +1,310 @@
...
+import urllib
+import random
+import warnings
+
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+from sklearn import preprocessing, grid_search
+from sklearn.cross_validation import train_test_split
```
dhimmel: Import from `sklearn.model_selection` so warning goes away.

dhimmel: See imports from [`2.TCGA-MLexample.ipynb`](https://github.com/cognoma/machine-learning/blob/7406e51f087932062a076dfbd5573c0d0fd8cbc2/2.TCGA-MLexample.ipynb) for how to rid these warnings.

algorithms/scripts/WIP-passiveaggressive-KT12.py
```diff
@@ -0,0 +1,310 @@
...
+n_feature_kept = 500
+
+param_fixed = {
+    'fit_intercept': True,
+    'random_state' : 0
+}
+
+param_grid = {
+    'loss'  : ['hinge', 'squared_hinge'],
+    'C': [10 ** x for x in range(-5, 5)]
```
dhimmel: Given the performance visualization, Would it make sense to focus in on the 10e-5 to 1 space for C? Same number of Cs but more concentrated around the sweet spots.

[See how we compute space in cognoml](https://github.com/cognoma/cognoml/blob/33045e853e57e18c06f64f6ab499da3081080a5a/cognoml/classifiers/logistic_regression.py#L19).

dhimmel (commented):
Really cool. Great algo name haha. The prediction distributions are beautiful with nice separation of positives and negatives.

Want to provide a little background of what the algo does, since most of us are probably not familiar with it?

algorithms/scripts/WIP-passiveaggressive-KT12.py
KT12: Thank you very much for your review.  It's a great name indeed.

The algorithm is passive (no update) when the example is classified correctly and the distance is > '1 unit' from the decision boundary.

The algorithm is aggressive (updates) when the classification is wrong or the classification is correct, but the distance is < '1 unit' from the decision boundary.   Loss function is hinge or hinge^2.  The update is proportional to the loss function (instead of using a learning rate).

It's similar to the SVM in that examples close to the boundary affect the decision boundary far more than examples very far from the boundary.  It's similar in update logic logic to the perceptron, but you can regularize a passive aggressive learner.

The learning time is very fast (can be used 'online').  It is sensitive to noise, be the feature set has been truncated in Cognoma's case.

This [slide deck](http://sssslide.com/speakerdeck.com/sorami/perceptron-support-vector-machine-and-passive-aggressive-algorithm) has a good overview of the passive aggressive algo compared to both the SVM and perceptron algos.

This lecture [video](https://www.youtube.com/watch?v=uxGDwyPWNkU) has a pretty good explanation.

KT12:
I have updated the method for importing packages and expanded the grid search for 'C' values in the passive aggressive classifier.

For the MLP, should I continue to troubleshoot?  Thank you.

KT12:
The Passive Aggressive Classifier is ready.

I completely removed the multilayer perceptron (for now).  Maybe I'll come back to it later.

Had to resort to `git revert --hard` to get my remote and master inline with each other.

dhimmel (approved):

dhimmel:
Really nice job cleaning this pull request up. Unless you have any more changes you want to make, I'm going to go ahead and merge.

Also let me know if I'm on the right track with my thinking: [`PassiveAggressiveClassifier`](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.PassiveAggressiveClassifier.html) is fitting a linear model, just like when we use SGDClassifier. Although PassiveAggressiveClassifier doesn't support `predict_proba`, it does support `decision_function` so we can get ranked predictions. Like elastic net logistic regression using SGDClassifier, we can get feature weights, which help us understand which features are influential. The main advantage compared to SGDClassifier is that PassiveAggressiveClassifier models are much quicker to train.

KT12:
Thank you for reviewing.  I think your intuition about the PA classifier is correct.

I thought I had previously posted [this video](https://www.youtube.com/watch?v=TJU8NfDdqNQ) and [slide deck](http://sssslide.com/speakerdeck.com/sorami/perceptron-support-vector-machine-and-passive-aggressive-algorithm), but they provide a good overview.

Issue #70, opened 2016-11-19
Memory issue


htcai:
I am running my notebook obtained by revising the latest _2.TCGA-MLexample_ in Ubuntu on my laptop (8GB RAM & 8GB swap). I used over-sampling which increased the size of the training data by about 7%. My machine keeps running into memory problem: `OSError: [Errno 12] Cannot allocate memory`, as well as other exceptions

There is no problem after I discard `pipeline`. I will use my MacBook (using compressed memory) to run the notebook, but it will be much slower.

dhimmel:
Okay I think this memory issue probably started after we merged https://github.com/cognoma/machine-learning/pull/54. It may be worth considering reverting our pipeline to the old incorrect ordering.

htcai:
@dhimmel Thanks for your reply! In the older version of pipeline, `k` is fixed in the pipeline while we feed a singleton list (e.g., [2000]) to the pipeline in the current version. Does this lead to the difference that GridSearchCV will fit `SelectKBest` only once in the former case while it will be fitted for each training fold in the latter?

dhimmel:
> Does this lead to the difference that GridSearchCV will run SelectKBest only once in the former case while it will be run for each training fold in the latter?

Previously the grid_search only included the SGDClassifier. Now the grid_search includes the entire pipeline. Therefore, cross validation now refits the feature selection (if used) and standardization (if used) on every training-fold rather than once on the entire X_train.

htcai:
Daniel, thank you for your confirmation! This information is very helpful.

KT12:
There is a fix suggested [here](http://scikit-learn.org/stable/faq.html#why-do-i-sometime-get-a-crash-freeze-with-n-jobs-1-under-osx-or-linux).

After implementing the fix, I tried using Isomap to do some dimensionality reduction but my Jupyter Notebook still yielded `OSError: [Errno 12] Cannot allocate memory`.

dhimmel:
@KT12 So you kept `n_jobs=1` in [`sklearn.manifold.Isomap`](http://scikit-learn.org/stable/modules/generated/sklearn.manifold.Isomap.html)? It's possible that even with only one job running `Isomap`, you could run out of memory.

You can also set `n_jobs=1` in GridSearchCV, not sure exactly what stage is causing you to run out of memory. See https://github.com/cognoma/machine-learning/issues/43#issuecomment-245379854 for more information on memory usage at different stages of our pipeline.

KT12:
[...]

PR #71, opened 2016-12-02
Improved the performance of the SGD classifier on sparse mutations by reducing the noise

htcai:
In response to [#52](https://github.com/cognoma/machine-learning/pull/52), I experimented a variety of methods to improve the performance of the SGD Classifier on sparse genes (e.g., RIT1, MAP2K1).

Mainly, I used PCA to reduce the noise and then employed LDA to compress the data to 2 dimensions. This strategy is justified by the apparent overfitting regarding many genes captured in [#52](https://github.com/cognoma/machine-learning/pull/52).

Taking RIT1 as the example, the improvement is notable. The test AUROC is 0.837, comparing with the 0.65 in [#52](https://github.com/cognoma/machine-learning/pull/52). For several other genes I explored (e.g., MAP2K1, SMAD2), the improvement is mostly between 0.07 and 0.10. Moreover, the total time needed for dimensionality reduction and model training is shorter than half a minute in Ubuntu on my ASUS laptop.

Still, there are a few issues. The optimal `n_component` of PCA for different genes is different. Therefore, it will be great if I can search over multiple values of `n_components`, which is probably viable using pipeline. However, Ubuntu (8GB RAM + 8GB swap) always runs out of memory when running the pipeline. My MacBook does not have this issue since it uses compressing memory, but it never ends running the pipeline. Hence, I didn't use pipeline in the file submitted.

Also, I am not sure how to interpret the probability plot at the end of the notebook.

Any suggestions and/or comments are welcome. Thanks in advance!


algorithms/scripts/RIT1-PCA+LDA-htcai.py
```diff
@@ -0,0 +1,219 @@
...
+# Here are the percentage of tumors with RIT1
+y.value_counts(True)
+
+
+# ## Set aside 10% of the data for testing
+
+# In[9]:
+
+# Typically, this can only be done where the number of mutations is large enough
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
```
gwaybio: Looks like with this split there are only `(6016 / 0.002874) * 0.1 = 1.7` gold standard positives in the test set. We may want to get a bit creative to overcome this sample size issue.

algorithms/scripts/RIT1-PCA+LDA-htcai.py
```diff
@@ -0,0 +1,219 @@
...
+'Size: {:,} features, {:,} training samples, {:,} testing samples'.format(len(X.columns), len(X_train), len(X_test))
+
+
+# ## Feature Standardization
+
+# In[10]:
+
+get_ipython().run_cell_magic('time', '', 'scale = StandardScaler()\nX_train_scale = scale.fit_transform(X_train)\nX_test_scale = scale.transform(X_test)')
```
gwaybio: any reason why you are fit_transforming `X_train` but not `X_test`? Same comment about PCA implementation below.

algorithms/scripts/RIT1-PCA+LDA-htcai.py
```diff
@@ -0,0 +1,219 @@
...
+# In[11]:
+
+get_ipython().run_cell_magic('time', '', 'n_components = 50\npca = PCA(n_components=n_components, random_state=0)\nX_train_pca = pca.fit_transform(X_train_scale)\nX_test_pca = pca.transform(X_test_scale)')
+
+
+# ## Dimensionality reduction via LDA
+
+# In[12]:
+
+get_ipython().run_cell_magic('time', '', '\nlda = LinearDiscriminantAnalysis(n_components=2)\nX_train_lda = lda.fit_transform(X_train_pca, y_train)\nX_test_lda = lda.transform(X_test_pca)')
```
gwaybio: wow! only 2 components! Can you add a scatter plot here where you color RIT1 mutation status?

algorithms/scripts/RIT1-PCA+LDA-htcai.py
```diff
@@ -0,0 +1,219 @@
...
+# Top predictions amongst negatives (potential hidden responders)
+predict_df.sort_values('decision_function', ascending=False).query("status == 0").head(10)
+
+
+# In[23]:
+
+# Ignore numpy warning caused by seaborn
+warnings.filterwarnings('ignore', 'using a non-integer number instead of an integer')
+
+ax = sns.distplot(predict_df.query("status == 0").decision_function, hist=False, label='Negatives')
```
gwaybio: It looks like the LDA transform causes the decision function to shift far to the negative. Any reason for this?

algorithms/scripts/RIT1-PCA+LDA-htcai.py
```diff
@@ -0,0 +1,219 @@
...
+ax = sns.distplot(predict_df.query("status == 1").decision_function, hist=False, label='Positives')
+
+
+# In[24]:
+
+ax = sns.distplot(predict_df.query("status == 0").probability, hist=False, label='Negatives')
+ax = sns.distplot(predict_df.query("status == 1").probability, hist=False, label='Positives')
```
gwaybio: It looks like the positives all have extremely low probability of being positive but the negatives are flat. Looks fishy. Is there an intuition as to why this is happening?

gwaybio (commented):
Nice work! Really cool idea combining PCA and LDA - it is surprising that the ROC curves look so good.

There are some plots (namely, the decision plot and probability plot) that are concerning. I made some comments inline.

I also have a general comment about RIT1:

1. It is an oncogene important in the RAS/MAPK pathway
2. Therefore, it would be interesting to test if a RIT1 classifier can also classify KRAS/NRAS/HRAS mutations.
3. Along the same lines, can a KRAS/NRAS/HRAS classifier predict RIT1?

I think these questions could be an alternative way of improving the performance of a RIT specific classifier - but may not work for other genes with low prevalence.

htcai:
@gwaygenomics thanks a lot for so many important comments and questions! I will try to answer the questions one by one.

algorithms/scripts/RIT1-PCA+LDA-htcai.py
htcai: When we train the models for feature transformation and dimensionality reduction, I think I should only use the information from the train data. A sample is from [Raschka 2015](http://nbviewer.jupyter.org/github/rasbt/python-machine-learning-book/blob/master/code/ch05/ch05.ipynb).

KT12:
@htcai  I'll just leave this [link](http://stats.stackexchange.com/a/109810/131198) for reference since it contains discussion of PCA >> LDA for classification.

htcai:
@KT12 thank you for the link! That is actually where I learned about this strategy. But now I believe that I need to re-read the post and the linked articles.

algorithms/scripts/RIT1-PCA+LDA-htcai.py
htcai: I decided to change the size of the testing data to 0.2 and set `stratify=y`, which probably is able to make the testing score more reliable.

algorithms/scripts/RIT1-PCA+LDA-htcai.py
htcai: Sorry that I misunderstood the component number for LDA. According to the documentation of [`sklearn.discriminant_analysis.LinearDiscriminantAnalysis`](http://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html), `n_components` cannot be larger than n_classes-1. Therefore, LDA cannot provide a one-dimensional data, which is probably not very informative. But now I am using PCA only, and keeping a 30-dimensional data.

htcai:
Following @dhimmel 's suggestion, I am using only PCA to remove the noise and reduce the dimensionality to 30. I renamed the file to _RIT1-PCA-htcai.ipynb_. As a side note, I corrected my misunderstanding of LDA. Specifically, `n_components` cannot be larger than n_class-1, according to the [documentation](http://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html). Therefore, LDA can only produce a 1-dimensional data in this case, which is not very (intuitively) informative.

Firstly, I set `test_size=0.2` and `stratify=y`, since there are very few positive samples in the dataset, as @gwaygenomics pointed out.

Also, I scaled the data both before and after PCA. If without the second time of scaling, the range of `decision_function` is extremely large (e.g., > 50,000). Also, it actually improved the testing score. Now the testing score is 0.843, even better than the result from the earlier analysis involving LDA.

In addition, the plot of `predict_proba` looks bumpy, which may well be a consequence of `PCA`.

Please let me know what I should do to further improve the notebook. Thanks in advance!

Moreover, I am interested in @gwaygenomics 's comments on the the path. I would like to further explore the connection between RIT1 and KRAS/NRAS/HRAS.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
+
+# coding: utf-8
+
+# # Create a logistic regression model to predict RIT1 mutation from gene expression data in TCGA
+
+# In[1]:
+
+import os
+import urllib
```
dhimmel: Can remove `urllib` and `random` imports

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
...
+# In[7]:
+
+# The Series now holds RIT1 Mutation Status for each Sample
+y.head(6)
+
+
+# In[8]:
+
+# Here are the percentage of tumors with RIT1
+y.value_counts(True)
```
dhimmel: Can you also run `y.value_counts()`? There are so few positives I want to know the exact number... this is cool!

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+# ## Feature Standardization
+
+# In[10]:
+
+get_ipython().run_cell_magic('time', '', 'scale_pre = StandardScaler()\nX_train_scale = scale_pre.fit_transform(X_train)\nX_test_scale = scale_pre.transform(X_test)')
+
+
+# ## Reducing noise via PCA
```
dhimmel: Let's get rid of `%%time` in this section for better diff viewing. And the "Feature Standardization" as well.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
...
+# Typically, this can only be done where the number of mutations is large enough
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, stratify=y, random_state=0)
+'Size: {:,} features, {:,} training samples, {:,} testing samples'.format(len(X.columns), len(X_train), len(X_test))
+
+
+# ## Feature Standardization
+
+# In[10]:
+
+get_ipython().run_cell_magic('time', '', 'scale_pre = StandardScaler()\nX_train_scale = scale_pre.fit_transform(X_train)\nX_test_scale = scale_pre.transform(X_test)')
```
dhimmel: I think you can write this all in a pipeline, which will simplify the code and make the process more clear. See example in [`2.TCGA-MLexample.py`](https://github.com/cognoma/machine-learning/blob/1c27195c485fd623f30dbeec57808fba1c71c687/scripts/2.TCGA-MLexample.py#L108-L112).

dhimmel: We can discuss at the meetup, whether we should do cross-validation the right or wrong but faster way as discussed in https://github.com/cognoma/machine-learning/issues/70#issuecomment-265901947.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
...
+get_ipython().run_cell_magic('time', '', 'scale_post = StandardScaler()\nX_train_scale = scale_post.fit_transform(X_train_pca)\nX_test_scale = scale_post.transform(X_test_pca)')
+
+
+# ## Parameters
+
+# In[13]:
+
+param_grid = {
+    'alpha': [10 ** x for x in range(-4, 1)],
+    'l1_ratio': [0, 0.2, 0.5, 0.8, 1],
```
dhimmel: I'd more densely sample `alpha` and reduce the `l1_ratio` to `[0, 0.15]`. See https://github.com/cognoma/machine-learning/issues/56.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
...
+cv_result_df = pd.concat([
+    pd.DataFrame(cv.cv_results_),
+    pd.DataFrame.from_records(cv.cv_results_['params']),
+], axis='columns')
+cv_result_df.head(2)
+
+
+# In[17]:
+
+# Cross-validated performance heatmap
```
dhimmel, line 150: This figure shows how hard the problem is. We're struggling to find a nice optimum of parameters. Performance is all over the place.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,216 @@
...
+# In[10]:
+
+get_ipython().run_cell_magic('time', '', 'scale_pre = StandardScaler()\nX_train_scale = scale_pre.fit_transform(X_train)\nX_test_scale = scale_pre.transform(X_test)')
+
+
+# ## Reducing noise via PCA
+
+# In[11]:
+
+get_ipython().run_cell_magic('time', '', 'n_components = 30\npca = PCA(n_components=n_components, random_state=0)\nX_train_pca = pca.fit_transform(X_train_scale)\nX_test_pca = pca.transform(X_test_scale)')
```
dhimmel: Would be nice to know what percentage of variance 30 components are capturing? It seems like 30 could be removing some important signals.

dhimmel (commented):
Nice job with some really interesting and helpful results. Made some comments. We can discuss at meetup tomorrow if you can make it.

algorithms/RIT1-PCA-htcai.ipynb
```diff
@@ -0,0 +1,891 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "0.5996\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Percentage of preserved variance\n",
+    "print('{:.4}'.format(sum(pca.explained_variance_ratio_)))"
```
dhimmel: So 35 components captures ~60% of variance. Can we increase the number of components until we capture at least 90% of variance?

htcai:
I made the following main changes.

1. I increased the proportion of testing data to 30%, which leads to a more interpretable visualization of probability (i.e., the last plot). As @gwaygenomics pointed out, a small portion of negative samples were mistakenly predicted to be positive, which is displayed as a second (but small) peak in the high-probability range. In addition, although this strategy slightly reduces the training and testing score (to 0.763), the testing score is still better than the baseline SGD Classifier (0.65).

2. Following @dhimmel 's suggestion, I fixed the `l1_ratio` and searched within a more condensed range of `alpha` (`np.linspace(10**-4, 10**-2, num=16)`). The validation results display a complicated effect of `alpha`'s change.

3. I calculated the ratio of explained variance of the 35 dimensions preserved in PCA (5 more than it was), which turns out to be 0.5996. This appears to be quite low. However, if more dimensions are kept, over-fitting will occur immediately. Moreover, I am applying PCA to reduce the noise in the data. A low value of `explained_variance_ratio` might not be unexpected.

4. Raw counts of positive samples are provided for both the testing set and the entire dataset.

algorithms/RIT1-PCA-htcai.ipynb
htcai: As I mentioned below, increasing the dimensionality leads to over-fitting and extremely low testing score immediately. It appears that a large proportion of the variation in the dataset is probably noise.

dhimmel:
Great work with this pull request -- this is interesting stuff!

Can you re-export `RIT1-PCA-htcai.py`?

> The validation results display a complicated effect of `alpha`'s change.

Here's the plot:

perf-grid

I don't actually think the effect of `alpha` is complicated. Instead I think we have a really noisy estimate of performance. It is concerning that we're not able to identify an optimal `alpha` range. Instead, the max performing values are scattered. Yuck! The bright side is if we can crack how to deal with these low number of positive situations, we're doing something really powerful.

I think we may be able to decrease the noise in our performance measurements by playing with the `cv` parameter of [`GridSearchCV`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html). Let's start by trying out 10-fold crossvalidation (`cv=10`). If that still doesn't clarify things, we may need repeated cross validation, although we'll have to see how inconvenient that is in `sklearn` (very according to [this answer](http://stats.stackexchange.com/a/253630/74908) although I don't think it should actually be that hard). 

> I calculated the ratio of explained variance of the 35 dimensions preserved in PCA (5 more than it was), which turns out to be 0.5996. This appears to be quite low. However, if more dimensions are kept, over-fitting will occur immediately.

Interesting. I'm worried that for some mutations the ~40% of uncaptured variance will be important. But I agree it's more important to actually be able to fit a model than include all the relevant features.

dhimmel:
We could also use [`StratifiedShuffleSplit`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.StratifiedShuffleSplit.html) -- while not as elegant as repeated cross-validation, it should work.

On second thought, perhaps it's even more elegant!

htcai:
@dhimmel Thanks a lot for your detailed and valuable comments! I will re-export the `.py` script after I further revise the notebook according to your suggestions. Unfortunately, pipeline still drain the memory even if I only cross-validate over two values of `alpha`.

In order to locate the optimal range of `alpha`, a cross-validation involving a larger number of values (e.g., 100, 200, etc) can be run. After the dimensionality reduction, it will probably take only less than one minute if `alpha` is the only hyper-parameter that is being tuned.

Training a classifier over an extremely imbalanced dataset has always been a thorny issue. I tried [under-sampling and over-sampling](https://github.com/scikit-learn-contrib/imbalanced-learn) several weeks ago, but it turned out to be a disaster. The performance is much worse than the baseline SGD classifier. I am happy to explore other methods later.

It is a good idea to try 10-fold cross-validation. My only concern is that there will be only 1 or 2 positive samples in each of the 10 folds. Still, only the experiment result can clarify whether that is a problem.

I agree that the removal of ~40% of the variance is a problem for other mutations. For the about 10 mutations I experimented, the optimal number of principal components is between 30 - 100. Therefore, if with sufficient computational resource, a GridSearchCV over `n_components` should work for our purpose of optimizing the classifier for a different mutation.

dhimmel:
> Therefore, if with sufficient computational resource, a GridSearchCV over n_components should work for our purpose of optimizing the classifier for a different mutation.

Great idea -- really insightful! Agree that computation time will be the drawback.

> My only concern is that there will be only 1 or 2 positive samples in each of the 10 folds.

There will be more positives in the training folds, which is ultimately more important. The larger number of testing folds (10 versus 3) will counterbalance the fewer number of positives in each. This is why leave one out CV is best, even though there are often 0 positives in a given testing fold.

htcai:
@dhimmel I used 10-fold GridSearchCV without `StratifiedShuffleSplit` and searched over a larger space of `alpha`: `[x for x in np.linspace(10**-4, 10**-2, num=200)]`. The performance measurement is still rather noisy. I probably need to try repeated cross-validation to reduce the noise.
image

On the other hand, `StratifiedShuffleSplit` tends to cause over-fitting and makes the selection prefer extremely large values.
image

dhimmel:
@htcai interesting. I'd love to see 1 × 10-fold for larger alphas, like you did for `StratifiedShuffleSplit`. I'm wondering if these higher alphas are actually producing better models... I don't really see a way that `StratifiedShuffleSplit` could cause inflated CV performance.

Do you want to switch the notebook to `StratifiedShuffleSplit`, so I can look over that code? Then let's see the coefficients of the best model to see what's going on with these high-performing large-alphas.

An alternative would be 10 × 10-fold CV, but that could be a pain to implement. And I don't really think it's fundamentally different from `StratifiedShuffleSplit` -- just a little more efficient.

htcai:
I updated the notebook with `StratifiedShuffleSplit` and searched for `alpha` over `[2**x for x in range(-20, 60)]`. Larger values of `alpha` are generally more preferable, but 0.5 is the optimal one.

It is good to see that the positive and the negative curves are more neatly segregated in the plots of `decision_function` and `probability`.

However, over-fitting seems to be salient.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -65,54 +63,78 @@
...
+# In[16]:
+
+sss = StratifiedShuffleSplit(n_splits=10, test_size=0.3, random_state=0)
+cv_folds = sss.split(X_train_scale, y_train)
+cv_folds = [[t, c] for t, c in cv_folds]
+
+
+# In[17]:
 
 get_ipython().run_cell_magic('time', '', "clf = SGDClassifier(random_state=0, class_weight='balanced', loss='log', penalty='elasticnet')\ncv = GridSearchCV(estimator=clf, param_grid=param_grid, n_jobs=-1, scoring='roc_auc')\ncv.fit(X = X_train_scale, y=y_train)")
```
dhimmel: Move everything besides `cv.fit(X = X_train_scale, y=y_train)` out of the `%%time` cell for better py export.

dhimmel: In `GridSearchCV` you need to specify `cv=sss`.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -65,54 +63,78 @@
...
-    'l1_ratio': [0, 0.2, 0.5, 0.8, 1],
+    'alpha': [2**x for x in range(-20, 60)],
+    'l1_ratio': [0]
 }
 
 
-# In[14]:
+# In[16]:
+
+sss = StratifiedShuffleSplit(n_splits=10, test_size=0.3, random_state=0)
```
dhimmel: If computation time doesn't become prohibitive, I'd love something more thorough such as:

```python
StratifiedShuffleSplit(n_splits=100, test_size=0.1, random_state=0)
```

`n_splits` times `test_size` equals the average number of times each observation will be used for testing in CV. The higher this is the less variation we should have in our performance estimates. However, we want `test_size` to be low, so the percent of observations used for training is maximized. Therefore, we should increase `n_splits` up until the point where it starts taking a long time. Make sense?

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -65,54 +63,78 @@
...
+cv_folds = sss.split(X_train_scale, y_train)
+cv_folds = [[t, c] for t, c in cv_folds]
```
dhimmel: Computing `cv_folds` yourself may not be necessary.

dhimmel (changes requested):

algorithms/scripts/RIT1-PCA-htcai.py
htcai: Yes, this is very sensible. It does not take a long time to run a 100-fold Grid Search. Actually, I set `n_split=150`. Now the heat map looks much less noisy.

dhimmel:
Cool one last thing so we can understand what's going on the super high alphas. Can you export the coefficients from the best model with `alpha = 2048`?

htcai:
That is a good idea! I visualized the coefficients of the `best_estimator_` with a bar plot.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,257 @@
...
+X_test_scale = scale_post.transform(X_test_pca)
+
+
+# ## Parameters and Classifier Fitting
+
+# In[15]:
+
+param_grid = {
+    'alpha': [2**x for x in range(-20, 60)],
+    'l1_ratio': [0]
```
dhimmel: Can add `0.15` to this list and rerun?

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,257 @@
...
+X_train_scale = scale_post.fit_transform(X_train_pca)
+X_test_scale = scale_post.transform(X_test_pca)
+
+
+# ## Parameters and Classifier Fitting
+
+# In[15]:
+
+param_grid = {
+    'alpha': [2**x for x in range(-20, 60)],
```
dhimmel: Let's reduce the upper bound to 30 at most.

dhimmel (commented):
I'm going over these results with @cgreene right now. We think the predicted probabilities should be centered around 0.0029 (`21 / (21 + 7285)`). So we're trying to diagnose what's going on. Two changes to `param_grid` will help us figure this out... hopefully.

htcai:
I searched over four values `[0, 0.05, 0.1, 0.15]` and the best parameters are `{'alpha': 1, 'l1_ratio': 0.05}`. Still, it seems that the predicted probabilities are higher than expected.

algorithms/scripts/RIT1-PCA-htcai.py
```diff
@@ -0,0 +1,257 @@
...
+plt.ylabel('True Positive Rate')
+plt.title('Predicting RIT1 mutation from gene expression (ROC curves)')
+plt.legend(loc='lower right');
+
+
+# ## Investigate the predictions
+
+# In[24]:
+
+X_transformed = scale_post.transform(pca.transform(scale_pre.transform(X)))
```
dhimmel: Did you want to put scale_pre, pca, and scale_post in a pipeline? This would really simplify things and help avoid errors.

dhimmel (commented):
Yeah I'm a little confused by the probabilities. Using a pipeline would help us make sure the right transformations are being applied.

Otherwise these are fantastic results. Really like the new CV plot!

htcai:
I finally got it run with pipeline (and hopefully of correct order) including scaling, PCA and classifier. Within a small amount of time, I can only search over a smaller space of `alpha` and use a 3-fold cross validation. As a consequence, I decided to increase `test_size` in `StratifiedShuffleSplit` to 0.2 and only use 20% of the entire data set for testing. Also, the CV plot does not seem as good as it was.

dhimmel:
Hey @htcai, I meant to suggest using a pipeline but preserving the exact functionality. Sorry I wasn't more clear. Feel free to revert to 544fd0322a1a4a88f4a3c56ab3e0878389ce1322.

I don't want to drag this pull request out, so either I can merge now, or we can talk Tuesday about the pipeline method I meant to suggest.

htcai:
@dhimmel Thanks for your clarification. I guess you are suggesting pipelining the scaling and the PCA while leaving the classifier alone. If this is not the case, we should talk further about the implementation tomorrow.

I increased the number of principal components to 50, which is a plausible value if we want a uniform one for all of the sparse genes. The testing score was actually improved for RIT1 comparing with 30 and 35.

dhimmel:
@htcai I remember that you made some progress last night at the meetup. Let me know when this is ready for me to take a look again.

htcai:
@dhimmel I am not sure whether you are referring to pipelining the data standardization and PCA. If this is the case, then it is already contained in the latest commit.

I tried `FastICA` and `NMF`. Both of them, with different `n_components`, either skew the prediction of probability (as does `PCA`) or drop the testing score to some level below random guess (i.e., 50%).

In addition, I am currently running `NMF` with 500 components. It seems that it will not finish within a reasonable amount of time.

dhimmel:
> @dhimmel I am not sure whether you are referring to pipelining the data standardization and PCA. If this is the case, then it is already contained in the latest commit.

Great, that's what I was referring to. This PR is ready to merge. Let's do a separate for the other dimensionality reduction methods.

dhimmel (approved):

PR #78, opened 2017-01-18
Isomap

KT12:
Used isomap to run non-linear dimensionality reduction before using logistic regression.  Because of the [computing resources needed](http://scikit-learn.org/stable/modules/manifold.html#complexity), the nearest neighbors used were 8 and the components retained were 64.  I ran it on a machine with 8GB RAM and 8GB Ubuntu swap.

The performance of the classifier would improve with higher settings, but would be difficult to execute on my machine.


dhimmel:
> The performance of the classifier would improve with higher settings, but would be difficult to execute on my machine.

It's certainly not performing that well with the current paramenters (: Good to know.

dhimmel (approved):

KT12:
Thank you for merging.

At least we now know that the gene expression features do not lie on a curved manifold and  linear dimensionality reduction will get good, computing resource-efficient classification.

PR #80, opened 2017-02-08
Multi-task classification

brankaj:
This notebook implements finding accuracy scores for genes that have targeted therapies. Since some agents target several genes, it is interesting to see can building an ensemble model improve results over a model that is built separately for each gene. Several methods support multi-task classification including Random forest. Multi-task version of Lasso also exists. These two methods are implemented in this notebook. Methods are evaluated using accuracy and precision scores.

Unfortunately, I am not sure does this notebook have enough findings to be a pull request. The reasons are as follows:

1. Lasso and Multi-task lasso do not provide difference in results.
2. Multi-task random forest provides improvement for some genes. For example, in the case of agent Ponatinib, there was a 1% improvement in accuracy score for genes KDR and ABL.
3. All methods seem to fail to recognize true positives. This is the reason why precision scores are 0 and this part of code is in the comment section. I hoped that by using different class weights, this problem would be solved. However, I could not find any weights that were successful in recognizing more samples with label '1'.


dhimmel:
> Several methods support multi-task classification including Random forest. Multi-task version of Lasso also exists. These two methods are implemented in this notebook. Methods are evaluated using accuracy and precision scores.

Cool!

> Unfortunately, I am not sure does this notebook have enough findings to be a pull request.

Always enough to be a pull request! I'll take a look.

dhimmel:
Nice to see the [`MultiTaskLasso`](http://scikit-learn.org/0.18/modules/generated/sklearn.linear_model.MultiTaskLasso.html) classifier in use.

> All methods seem to fail to recognize true positives.

Your intuition is spot on. It looks like every sample is being classified as a negative. Hence, accuracy is just the proportion of samples that are negative. I'm guessing your balanced accuracy would be 0.5 and kappa would be 0 (i.e. no predictive ability). I think part of the problem could be that you're not trying different `alpha` (regularization strength) values. Usually, we use `GridSearchCV` to try a range of `alpha` values. While the `GridSearchCV` will make things slower, I think optimizing alpha is the place to start.

explore/MultiTaskClassification.py
```diff
@@ -0,0 +1,461 @@
...
+
+Classifiers_mono = {
+    "Random forest": RandomForestClassifier(max_depth=100, min_samples_split=2, n_estimators=10, random_state=0,class_weight = weights),
+    "Lasso": Lasso(alpha=0.1),
+}
+
+
+# In[7]:
+
+# Agents and targeted genes
```
dhimmel: These dictionaries are really useful. Thanks for doing this reconnaissance work. You may want to consider making a TSV file for this data, so you don't have to code it in the notebook. I could imagine a table with three columns: entrez_gene_id, gene_symbol, and drug.

brankaj: I will first try different values of alpha. TSV file is a good idea, it will make this notebook better organized. Thank you!

explore/MultiTaskClassification.py
```diff
@@ -0,0 +1,461 @@
...
+
+for GENE in list(genes.keys()):
+    
+    y_train_each = y_train[GENE]
+    true_label = list(y_test[GENE])
+    
+    y_test_predict = dict()
+    for name, classifier in Classifiers_mono.items():
+        classifier.fit(features_train, y_train_each)
+        y_test_predict[name] = classifier.predict(features_test)
```
gwaybio, line 159: I am a big fan of this analysis. Will be good to eventually get a sense of _how_ these therapeutics act. Do they target the mutation or the downstream consequences of the mutation?

> It looks like every sample is being classified as a negative. Hence, accuracy is just the proportion of samples that are negative. 

Alternatively, could it be that the pipeline is fit_transformed for different X matrices (line 247)(resulting in different features for training and testing) and then fit using training features and predicting using testing features?

dhimmel: Ah @gwaygenomics, really great catch. @brankaj in the following line, you should just be applying the transformation, not refitting it:

```python
features_test=pipeline.fit_transform(X=X_test, y=y_test)
```

brankaj: Thank you guys so much for your comments. I am really sorry for this huge delay; finishing up my dissertation is not leaving me enough time to do other things. 
Regarding new files, I corrected the fit-transform issue and I added AUROC values and CV(cross validation) form of classifiers. Also, I made csv file with drug and gene information so someone can use it for future research. 
Unfortunately, I am not happy with the ROC curves and I am not sure how to improve them. If you think it is not appropriate to leave this analysis in this condition, please let me know and I will delete these files.

explore/MultiTaskClassification.py
```diff
@@ -0,0 +1,308 @@
...
+        
+    
+    result_RF = y_test_predict['Random forest']
+    result_L = binarize(y_test_predict['LassoCV'].reshape(-1,1), threshold=0.5)     
+    as_RF.append(accuracy_score(true_label,result_RF))
+    as_L.append(accuracy_score(true_label,result_L))
```
dhimmel, line 165: Accuracy score is not a good performance metric here, since our data is not balanced (we have more negatives than positives). For example, you can predict every sample as negative and get near 100% accuracy. You can use balanced accuracy instead. Or you can use AUROC, which I recommend since it accounts for multiple classification thresholds not just a single threshold.

explore/Target_therapy.csv
```diff
@@ -0,0 +1,47 @@
+Drug,Entrez_gene_id,Gene_symbol
```
dhimmel, line 1: can we remove capitalization of first letter? Specifically, `entrez_gene_id` is the column name we use in `cognoma/genes` and `cognoma/cancer-data`.

dhimmel (commented):
Hey @brankaj this is exciting work. Sorry it took me so long to get around to reviewing it.

It's difficult to evaluate until we fix the accuracy / balance issue described in the my comments.

Thanks!

explore/Target_therapy.csv
brankaj: Ok, I will change the file and also change or remove the accuracy score. I have added AUROC values, but the results are not encouraging. I will first try the balanced accuracy scores. Hopefully, this approach will work. 
Thanks.

dhimmel: > I have added AUROC values, but the results are not encouraging.

That's fine! Will still be helpful to see AUROC.

PR #81, opened 2017-02-13
Isomap with more dimensions

KT12:
Re-ran the same analysis with Isomap dimensionality reduction, but with a larger number of components retained.  Results were similarly average.


algorithms/scripts/SGDClassifier-Isomap-KT12.py
```diff
@@ -3,7 +3,7 @@
 
 # # Create a logistic regression model to predict TP53 mutation from gene expression data in TCGA using Isomap as a dimensionality reduction technique
 
-# In[1]:
+# In[27]:
```
dhimmel: Ah the diff would be way nicer with a freshly run notebook (so cells start at 1). Is is feasible to restart and run all cells, so this part of the diff goes away?

dhimmel (commented):
So this PR increases in the number of components from 64 to 256 and adds parallel computation. Did performance change at all?

algorithms/scripts/SGDClassifier-Isomap-KT12.py
KT12: Sorry about that - must have made another change and re-run the notebook.

Test AUROC went from ~63% to ~75%.

dhimmel: > Test AUROC went from ~63% to ~75%.

Wow! So it was in part an issue of too few features.

KT12:
Unfortunately, there is no random seed to set for Isomap dimensionality reduction.  This time the AUROC was ~ 70%.

dhimmel:
> Unfortunately, there is no random seed to set for Isomap dimensionality reduction.

Wow, that is a major drawback.

PR looks good.

dhimmel (approved):

PR #84, opened 2017-02-14
Create Vega specification for AUROC bar plots

patrick-miller:
This introduces a Vega based AUROC bar plot, used for comparing models on different mutations.

image

I'm thinking we maybe add a dashed line for the test partition performance of the covariates-only model on each of these to give a sense for a baseline?


dhimmel:
@patrick-miller really nice! Looks like the test/train color scheme is similar to the [ROC curve plot](https://github.com/cognoma/machine-learning/pull/77#issuecomment-276768553), which is great -- let's make the colors exactly the same if possible!

> I'm thinking we maybe add a dashed line for the test partition performance of the covariates-only model on each of these to give a sense for a baseline?

Yeah, I think a dashed line could work. IIn some cases the covariate AUROC could be greater than the full AUROC. Accordingly, I think this is actually a difficult visualization that may take some trial and error to see what works. So I'd try out what you think will communicate this effectively!

patrick-miller:
Here is my first attempt at adding covariate only results:

image

The color scheme was the same, I had just added some transparency. I included the same transparency here.

dhimmel:
Nice. I think we should show the covariate performance for both testing and training. Is it possible to split the red dash into two? You could even use the same color scheme (orange / purple) for the dashses without the transparency.

patrick-miller:
Cool, how does this look -- is it clear enough?

image

dhimmel (approved):
@patrick-miller that's great. I like how you can see the covariate performance, but it's not visually overwhelming.

PR #92, opened 2017-04-13
Rename total column

George-Zipperlen:
Quick fix renaming 'TOTAL' column to '_Total' to visually distinguish it from gene-symbols, 
and force it to sort after the genes,


George-Zipperlen:
I've merged this branch back into the 'fix-heatmap-labels' branch on my local copy,
and pushed that upstream, so this branch is no longer needed

PR #99, opened 2017-06-06
Update Vega plot in notebook

patrick-miller:
So far this PR has made it so the Vega plot doesn't save intermediate files.

I am exploring how to work https://github.com/altair-viz/jupyter_vega into the notebook. I think that will make the plots dynamic in a pre-run notebook.

Issue #97, opened 2017-05-31
Standardize the plots in notebooks


patrick-miller:
I wanted to open this issue so that we could discuss what framework we wanted to use in the notebook that we export for our users. There are a lot of [options](https://www.youtube.com/watch?v=FytuB8nFHPQ). 

Right now, the majority of the plots are in **matplotlib/seaborn**. There is a pretty good replication of R's ggplot2 with yhat's **[ggpy](https://github.com/yhat/ggpy)**.

We can use JavaScript based, dynamic plots with **bokeh** or **Vega**. There has already been some work done with Vega (#74, #77, #84), and these can be incorporated into notebooks. The benefit of using Vega is that once we build a frontend results viewer, presumably we will be moving towards this dynamic method of displaying plots. Maintaining consistency would be a plus.

One issue with displaying Vega plots with ipyvega, is that (as detailed in this [ipyvega pull request](https://github.com/vega/ipyvega/pull/32)) the plots are exported to a static image when viewed through GitHub or nbviewer. Only live notebooks get the JavaScript version. As we are going to be serving a static notebook, I assume we would have the same result. I'll be looking into if [this new project](https://github.com/altair-viz/jupyter_vega) holds any answers.

George-Zipperlen:
Thanks @patrick-miller for the pointer to https://github.com/altair-viz/jupyter_vega.

I will look into this.

I had reached a brick wall with altair/vega-lite in that some of the graph attributes were not accessible in the available json properties / python classes.

For example horizontal / vertical orientation of axis and legend.  The examples force a horizontal orientation to the heatmap legend (quantitative scale), and a vertical orientation to a scatterplot legend with nominal scale.  If I tried using 'vertical' or 'horizontal' in a legend definition, it just wouldn't work.

dhimmel:
@patrick-miller I think your comment does a good job summarizing the options. I wish I were more familiar with the options. I've never used bokeh or ggpy, but they both seem like promising options.

I do think it would be nice to get away from seaborn/matplotlib. Seaborn is nice, but has very limited and predefined functionality. And tweaking the raw matplotlib is a pain. But obviously we'd need a better alternative. So everyone feel free to play around and provide feedback on which solution you think is best.

George-Zipperlen:
I propose using Bokeh instead of Vega-Lite for Cognoma visualizations:

+ Bokeh supports interactive charts
+ Grammar of Graphics influenced
+ abstract rendering
+ dynamic downsampling
[...]

Issue #94, opened 2017-05-26
Evaluate dask-searchcv to speed up GridSearchCV


dhimmel:
I'm excited about trying out dask-searchcv as a drop-in replacement for `GridSearchCV`. For info on dask-searchcv see, the [blog post](http://jcrist.github.io/introducing-dask-searchcv.html), [github](https://github.com/dask/dask-searchcv), [docs](http://dask-searchcv.readthedocs.io/en/latest/), and [video](https://youtu.be/RA_2qdipVng?t=35m08s).

I'm hoping using dask-searchcv for GridSearchCV will help solve the following problems:

1. High memory usage, e.g. https://github.com/cognoma/machine-learning/issues/70, caused by joblib overhead.
2. The slow performance of the pipeline when properly implementing cross-validation. See discussion at https://github.com/scikit-learn/scikit-learn/issues/7536#issuecomment-250841108. The builtin GridSearchCV is repeating the same transform steps making it brutally slow.

I initially mentioned dask-searchcv in https://github.com/cognoma/machine-learning/pull/93#issuecomment-302851217, a PR by @patrick-miller. I thought this would be a good issue for @rdvelazquez to work on. @rdvelazquez are you interested?

We'll have to add some additional dependencies to our environment. It may be a good time to also update the package versions of existing packages (especially pandas).

rdvelazquez:
Yea, I'm definitely interested! I'll have a little time to look into this before next Tuesday's meet-up so I'll come ready with some questions and suggestions to discuss.

@patrick-miller : Did you look into dask-searchcv yet? I don't want to duplicate work.

patrick-miller:
I only read the docs, so go for it! 

It isn't directly relevant to dash-searchcv, but one thing we need to keep in mind when implementing PCA on our features is that we want to only run PCA on the expression features and not the covariates. I have created a new issue for just this #96.

rdvelazquez:
I'm still looking into this and hope to post a WIP pull request early next week so you can see what I'm working on. So far dask-searchcv itself is very easy to implement but it does not seem to negate all the issues associated with running PCA in the pipeline... I'll have more soon.

rdvelazquez:
Here's my WIP notebook: [dask-searchCV](https://github.com/rdvelazquez/machine-learning/blob/daskSearchCV/explore/dask-searchCV/dask-searchCV.ipynb)

@dhimmel and @patrick-miller, some questions on where to go from here:
[...]

PR #100, opened 2017-06-06
Add PCA on expressions only to the CV pipeline

### LINKED ISSUE #96: PCA (and other pre-processing steps) on just the expression matrix in CV pipeline
We have two sources of features: the covariates and the gene expression matrix. When pre-processing this data, we generally want to perform dimensionality reduction only on the expression matrix.

This can prove cumbersome when trying to implement PCA in a pipeline. Scikit-learn does not provide this sort of functionality out of the box, but I believe it is possible to continue using pipelines. [Here is an example.](https://stackoverflow.com/questions/24124622/pipeline-with-pca-on-feature-subset-only-in-scikit-learn#24125176)

patrick-miller:
Closes #96

This includes PCA in the cross-validation pipeline so that we aren't leaking information to the validation set. It currently uses `FeatureUnion` which seems to work just fine.

We can add number of components into the grid search if we want pretty easily. Right now, it just has 100 components hard coded.


rdvelazquez:
Very nice work @patrick-miller ! I like the use of `FunctionTransform`. 

I only reviewed the pipeline section of this notebook. My only question is, are you normalizing the PCA components? I think our previous implementations ran a pipeline with (StandardScaler, PCA, StandardScaler) and it looks like your pipeline removed the post-PCA StandardScaler. I don't know if this will have implications on performance (it may just effect the speed of the classifier) but I thought I would ask the question. 

Also, you may want to move the .py file to the scripts folder.

patrick-miller:
Thanks! The `FunctionTransform` idea was not mine. The .py file has been moved.

As for normalizing the PCA components, I generally do so beforehand unless there is a good reason (perform PCA on correlation matrix instead of covariances). However, I have never scaled the resulting components before. I can certainly add it back in, maybe @dhimmel has some thoughts.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -281,26 +284,33 @@ def get_threshold_metrics(y_true, y_pred):
...
-# In[22]:
-
-def get_coefficients(classifier, X_mat):
-    coef_df = pd.DataFrame.from_items([
-        ('feature', X_mat.columns),
-        ('weight', classifier.coef_[0]),
-    ])
+# In[19]:
 
+def get_coefficients(classifier, feature_set):
```
dhimmel: Perhaps we should move this function to `utils.py`? This functionality should really be in `sklearn`... not sure most users really need to see the details.

dhimmel:
> However, I have never scaled the resulting components before. 

I'm pretty sure that standardization after PCA is not necessary, since the features will come out of PCA essentially standardized. @patrick-miller @rdvelazquez is this true? It probably won't hurt to standardize after PCA, but for efficiency we should avoid this unless it's actually needed.

dhimmel:
@patrick-miller did you want to fix the conflicts that arose from #99 and #102?

Merging conflicts in the notebook will likely be unpleasant. [`nbdime`](https://github.com/jupyter/nbdime) may be able to help.

It's also possible it may be easier to just manually modify the notebook to the desired state without using git.

patrick-miller:
The features coming out of PCA don't have unit variance, but they are on the same scale if you normalize beforehand (i.e. there are no orders of magnitude differences that aren't meaningful).

Sure, I will fix it up, don't think it should be too bad.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: We could move it. I'm not sure if there is a better way to extract these, or if we even should be doing it this way. The only reason I see for not moving it is if the researchers wants to change the dimensionality reduction, the function as-is will break.

I think what we probably want here is an inverse transformation on the pipeline to find out which genes have the highest loadings on the components that matter?

rdvelazquez:
> The features coming out of PCA don't have unit variance, but they are on the same scale if you normalize beforehand (i.e. there are no orders of magnitude differences that aren't meaningful).

Makes sense to me. Thanks for the explanation!

scripts/2.TCGA-MLexample-covariates.py
dhimmel: @patrick-miller IMO it's a real sklearn deficiency / bug to not be able to extract labels and provenance info for each feature: see https://github.com/scikit-learn/scikit-learn/issues/7536.

We created a [workaround here](https://github.com/cognoma/cognoml/blob/f9669edfe0871cd161ffefa3409f67c98dd407a2/cognoml/utils.py#L62-L94). I'm not sure if it will work on your more advanced pipeline, but worth a try.

 > The only reason I see for not moving it is if the researchers wants to change the dimensionality reduction, the function as-is will break.

I don't think it's realistic for most users to be able to change this, since it requires quite intricate knowledge of sklearn. I say we aim for a robust workaround, but if not possible we can use some hacks. Also if we can detect the failure cases, we can just return feature names like `X1`, `X2`...

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
-
-y_train = y[train_index]
-y_test = y[test_index]
-
-'Size: {:,} features, {:,} training samples, {:,} testing samples'.format(
-    len(X_partitions['full']['train'].columns), 
-    len(X_partitions['full']['train']), 
-    len(X_partitions['full']['test']))
+# Select the feature set for the different models
+N_COVARIATES = covariates.shape[1]
```
dhimmel: `covariates` is a pandas dataframe here. `len(covariates.columns)` preferred for readability.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
+def select_feature_set_columns(X, feature_set):
+    if feature_set=='expressions': return X[:, N_COVARIATES:]
```
dhimmel: newline before return statement.

Let's do `X.iloc[:, N_COVARIATES:]` to make it more explicit that we're doing integer slicing.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
-    'classify__l1_ratio': [0, 0.2, 0.8, 1],
+n_components_list = [50, 100]
+
+param_grids = {
+    'full': {
+        'features__expressions__pca__n_components' : n_components_list,
+        'classify__loss': ['log'],
+        'classify__penalty': ['elasticnet'],
+        'classify__alpha': [10 ** x for x in range(-3, 1)],
+        'classify__l1_ratio': [0, 0.2, 0.8, 1],
```
dhimmel: Can we reduce to fewer `l1_ratio` values? Perhaps even just a single value (default of 0.15%) as discussed in #56. @patrick-miller this extra parameter will add lot's of runtime for little benefit (unless the user wants to specify it themselves).

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
 
 # Parameter Sweep for Hyperparameters
-param_grid = {
-    'classify__loss': ['log'],
-    'classify__penalty': ['elasticnet'],
-    'classify__alpha': [10 ** x for x in range(-3, 1)],
-    'classify__l1_ratio': [0, 0.2, 0.8, 1],
+n_components_list = [50, 100]
+
+param_grids = {
```
dhimmel: really elegant here!

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
+    'full': {
+        'features__expressions__pca__n_components' : n_components_list,
+        'classify__loss': ['log'],
```
dhimmel: Since `log` and `penalty` are always the same do you want to add those to the three dictionaries in a loop rather than setting it 3 times. Looking to simplify & shorten code. `dict.update(common_params)` will make this clean.

scripts/2.TCGA-MLexample-covariates.py
dhimmel: Let's move this function to be under "Set aside 10% of the data for testing"

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
+    'covariates': {
+        'classify__loss': ['log'],
+        'classify__penalty': ['elasticnet'],
+        'classify__alpha': [10 ** x for x in range(-3, 1)],
+        'classify__l1_ratio': [0, 0.2, 0.8, 1],
+    }
 }
 
-pipeline = Pipeline(steps=[
+# Feature selection
```
dhimmel: Let's split "Feature selection" into its own cell for visual clarity.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
 ])
 
+# Full model pipelines
+pipeline_definitions = {
+    'full': Pipeline([
+        ('features', FeatureUnion([
+            ('expressions', expression_features),
+            ('covariates', covariate_features)
+        ])),
+        ('classify', SGDClassifier(random_state=0, class_weight='balanced'))
```
dhimmel: Let's define these once above:

```python
sgd_classifier_params = {
    'random_state': 0,
    'class_weight': 'balanced',
}

... SGDClassifier(**sgd_classifier_params)
```

Could even put the invariant `loss` and `penalty` here (unless there is some `sklearn` bug.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
+        ('features', FeatureUnion([('covariates', covariate_features)])),
+        ('classify', SGDClassifier(random_state=0, class_weight='balanced'))
+    ])
+}
+
 models = ['full', 'expressions', 'covariates']
 
 cv_pipelines = {mod: GridSearchCV(estimator=pipeline, 
-                             param_grid=param_grid, 
+                             param_grid=param_grids[mod],
```
dhimmel: indent to where estimator is

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
-                             scoring='roc_auc') for mod in models}
+                             scoring='roc_auc') 
+                for mod, pipeline in pipeline_definitions.items()}
 
 
-# In[15]:
+# In[12]:
 
-get_ipython().run_cell_magic('time', '', "for model, pipeline in cv_pipelines.items():\n    print('Fitting CV for model: {0}'.format(model))\n    pipeline.fit(X=X_partitions.get(model).get('train'), y=y_train)\n# cv_pipeline_full.fit(X=X_train_full, y=y_train)")
+get_ipython().run_cell_magic('time', '', "for model, pipeline in cv_pipelines.items():\n    print('Fitting CV for model: {0}'.format(model))\n    pipeline.fit(X=X_train, y=y_train)")
```
dhimmel: Would be great to remove `%%time` here and use https://docs.python.org/3/library/time.html#time.perf_counter.

This way we can get printouts for the time of each model and the ugly script export cell goes away in git diffs.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
...
-    # Best Params
     print(pipeline.best_params_)
 
 
 # ## Visualize hyperparameters performance
 
-# In[17]:
+# In[14]:
 
 cv_results_df_dict = {model:
```
dhimmel: I think this should be a single dataframe with a `model` column. Then you can do 
```
cv_results_df.query("@model == 'full'")
```

below (if we keep the grid viz cell below.)

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -234,7 +237,7 @@ def get_threshold_metrics(y_true, y_pred):
 }
 
 
-# In[20]:
+# In[17]:
```
dhimmel: Surprising to me that the `full`  (covariates + expression) performs almost equivalently to `expression` for TP53. Good to know. CC @gwaygenomics

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -280,28 +283,13 @@ def get_threshold_metrics(y_true, y_pred):
...
-        ('feature', X_mat.columns),
-        ('weight', classifier.coef_[0]),
-    ])
-
-    coef_df['abs'] = coef_df['weight'].abs()
-    coef_df = coef_df.sort_values('abs', ascending=False)
-    
-    return coef_df
-
 coef_df_dict = {
```
dhimmel: Let's make this a single df with a `model` column. Same with prediction dataframe. Simplifies output threefold when we have to store it.

dhimmel (changes requested):
@patrick-miller great work!

I've done a very detailed review of the code, since this will be such an important notebook going forward. As such, @rdvelazquez your input is also very welcome.

Let me know if you disagree with any of the comments. I erred on the side of over-commenting!

dhimmel:
Confirming @patrick-miller that this closes https://github.com/cognoma/machine-learning/issues/96? If so, edit original comment to say `Closes #56` and GitHub will automatically close the issue upon merge.

scripts/2.TCGA-MLexample-covariates.py
dhimmel: Following snippet may be helpful technique:

```python
dfs = list()
# append to dfs
pandas.concat(dfs)
```

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: I think that it may get converted to a numpy array in the pipeline, but I'll test it out.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: Sounds good to me. It definitely will reduce the runtime by a significant amount.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: I'll take a look, the printouts have been really annoying.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: I like this solution.

scripts/2.TCGA-MLexample-covariates.py
dhimmel: > I think that it may get converted to a numpy array in the pipeline

Yes I think you're likely right. If so, just moving it closer to it's usage will be helpful.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -80,108 +79,112 @@
 y.value_counts(True)
```
rdvelazquez: This should be percentage of tumors with _TP53_ right?

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: With L1 at 0.15, it looks like we are losing a good bit of the performance separation between expressions and covariates.

dhimmel: Interesting. How much? Will also need to see the performance at tested `alpha` values to make sure we're hitting its sweet spot.

patrick-miller: Covariates stayed at ~84%, while the others went from 91% to 87%. As I would generally expect, `L1 ratio` of 0 performed best for all models. In the full and expressions models, `alpha` is at the upper bound of 1, so there is likely room to improve.

rdvelazquez (commented):
Once again @patrick-miller this is great! Thanks for letting me provide some comments. I had no substantial revisions but I think it's always good to have a second pair of eyes looking at things. Good luck with the revisions. Excited to see this merged!

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: This is the case. I think it is clearer what is going on with the re-ordering of the sections in the new notebook.

scripts/2.TCGA-MLexample-covariates.py
rdvelazquez: Are you seeing any of my other comments? or is this the only one that got saved?

patrick-miller: This is the only one I see, I think. Maybe something happened when I was pushing a commit?

dhimmel: @rdvelazquez perhaps you have commits pending as part of a review? Refresh page?

rdvelazquez: I had some other comments as well, nothing too substantial, but I'm not seeing them anywhere in this thread... not sure what may have happened. @patrick-miller if you don't see any of my other comments, let me know and I'll re-add them tomorrow morning.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -6,6 +6,7 @@
 # In[1]:
 
 import os
+import time
 import urllib
```
dhimmel: Good time to remove unused imports. e.g. `urllib` and `random`

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -90,55 +91,64 @@
...
 
-# Select the feature set for the different models
-N_COVARIATES = covariates.shape[1]
+
+# ## Feature selection
+
+# In[11]:
+
+# Select the feature set for the different models within the pipeline
+N_COVARIATES = len(covariates.columns)
```
dhimmel: no reason for n_covariates to be ALL CAPS as per python conventions.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -90,55 +91,64 @@
...
     },
     'expressions': {
         'features__expressions__pca__n_components' : n_components_list,
-        'classify__loss': ['log'],
-        'classify__penalty': ['elasticnet'],
-        'classify__alpha': [10 ** x for x in range(-3, 1)],
-        'classify__l1_ratio': [0, 0.2, 0.8, 1],
+        'classify__alpha': regularization_alpha_list
     },
     'covariates': {
```
dhimmel: Does it break the grid search to add a step that isn't in the pipeline? If not, the same `param_grid` could be used for all three models?

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -147,33 +157,47 @@ def select_feature_set_columns(X, feature_set):
...
+st = time.perf_counter()
 
+# Fit the models
+for model, pipeline in cv_pipelines.items():
+    print('Fitting CV for model: {0}'.format(model))
+    pipeline.fit(X=X_train, y=y_train)
+    
+et = time.perf_counter()
+
+def pretty_time(seconds):
```
dhimmel: No need to reimplement:

```python
str(datetime.timedelta(seconds=seconds))
```

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -184,36 +208,38 @@ def select_feature_set_columns(X, feature_set):
...
         pd.DataFrame(pipeline.cv_results_),
-        pd.DataFrame.from_records(pipeline.cv_results_['params']),
-    ], axis='columns') for model, pipeline in cv_pipelines.items()}
+        pd.DataFrame.from_records(pipeline.cv_results_['params'])
+    ], axis='columns')
+    df['feature_set'] = feature_set
+    return df
 
-model = 'full'
+cv_results_df = pd.concat([
```
dhimmel: I think a for loop would be more readable than list comprehension here, because you could get rid of the `build_cv_results_df` function. Since it would be inline.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -283,27 +309,31 @@ def get_threshold_metrics(y_true, y_pred):
...
+model_coef_df = coef_df[coef_df['feature_set'] == model]
+
 print('{:.1%} zero coefficients; {:,} negative and {:,} positive coefficients'.format(
-    (coef_df_dict[model].weight == 0).mean(),
-    (coef_df_dict[model].weight < 0).sum(),
-    (coef_df_dict[model].weight > 0).sum()
+    (model_coef_df.weight == 0).mean(),
+    (model_coef_df.weight < 0).sum(),
+    (model_coef_df.weight > 0).sum()
 ))
```
dhimmel: Doesn't make that much sense to sum these values across models. Consider a `pandas.crosstab` solution comparing sign to model.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: I tried that at first, and it broke. I will attempt it again and report what error I get.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: I suppose we aren't going over a day in run time (or does that work anyway?)

EDIT: it does!

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: Yup, good call

patrick-miller: How is this?

```
def zero(w): return '{:.1%}'.format(np.mean(w == 0))
def negative(w): return '{:,}'.format(np.sum(w < 0))
def positive(w): return '{:,}'.format(np.sum(w > 0))

coef_df.groupby('feature_set')['weight'].agg([zero, negative, positive])
```

image

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: Here is the error:
image

I'll play around with fixing this some more.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -20,12 +20,11 @@
 from sklearn.linear_model import SGDClassifier
```
rdvelazquez, line 18: Do we need to import numpy, matplotlib.pyplot, and/or preprocessing?

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -76,146 +75,166 @@
...
-'Size: {:,} features, {:,} training samples, {:,} testing samples'.format(
-    len(X_partitions['full']['train'].columns), 
-    len(X_partitions['full']['train']), 
-    len(X_partitions['full']['test']))
+# Classifier: Elastic Net
+classifier = SGDClassifier(penalty='elasticnet',
+                           l1_ratio=regularization_l1_ratio,
+                           loss='log', 
+                           class_weight='balanced',
+                           random_state=0)
```
rdvelazquez, line 150: Clean! This is really nice (and easy to follow).

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -76,146 +75,166 @@
...
+    ('select_features', FunctionTransformer(select_feature_set_columns,
+        kw_args={'feature_set': 'expressions'})),
+    ('standardize', StandardScaler()),
+    ('pca', PCA())
+])
+covariate_features = Pipeline([
+    ('select_features', FunctionTransformer(select_feature_set_columns,
+        kw_args={'feature_set': 'covariates'})),
+    ('standardize', StandardScaler())
+])
```
rdvelazquez, line 119: This is impressive!! I'm not sure I can follow exactly what's going on. It may be worth adding a comment or two to explain  the general idea of needing to perform PCA on only the expression data and how this is implemented.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -76,146 +75,166 @@
...
-explained_variance = pca.explained_variance_
-expression_pca = pca.transform(scaled_expression)
-expression_pca = pd.DataFrame(expression_pca)
-expression_pca = expression_pca.set_index(expression.index.values)
+X = pd.concat([covariates, expression], axis=1)
+print('Gene expression matrix shape: {0[0]}, {0[1]}'.format(expression.shape))
+print('Covariates matrix shape: {0[0]}, {0[1]}'.format(covariates.shape))
 
+# Typically, this can only be done where the number of mutations is large enough
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.1, random_state=0)
```
rdvelazquez: > Typically, this can only be done where the number of mutations is large enough

 This comment seems to imply that we wouldn't perform a test train split when the number of mutations is small. I think we would always be performing a test train split regardless of the number of mutations. We should either revise or remove this comment.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -76,146 +75,166 @@
...
 # In[9]:
 
-# Here are the percentage of tumors with NF1
+# Here are the percentage of tumors with TP53
 y.value_counts(True)
 
 
-# ## Pre-process data set
-# TODO: currently running PCA on both train and test partitions
+# ## Set aside 10% of the data for testing
```
rdvelazquez, line 81: May want to split this into two cells or revise the header.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -76,146 +75,166 @@
...
-    ], axis='columns') for model, pipeline in cv_pipelines.items()}
-
-model = 'full'
+        pd.DataFrame.from_records(pipeline.cv_results_['params'])
+    ], axis='columns')
+    df['feature_set'] = model
+    cv_results_df = cv_results_df.append(df)
 
-cv_results_df_dict[model].head(2)
+cv_results_df.head(2)
```
rdvelazquez: Do we need this output?

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -281,66 +300,52 @@ def get_threshold_metrics(y_true, y_pred):
...
 model = 'full'
-
-print('{:.1%} zero coefficients; {:,} negative and {:,} positive coefficients'.format(
-    (coef_df_dict[model].weight == 0).mean(),
-    (coef_df_dict[model].weight < 0).sum(),
-    (coef_df_dict[model].weight > 0).sum()
-))
-coef_df_dict[model].head(10)
+model_coef_df = coef_df[coef_df['feature_set'] == model]
+model_coef_df.head(10)
```
rdvelazquez: Just as an FYI - An improvement for the future could be trying to understand which genes have the most impact. (the PCA components are not very meaningful). We could use `inverse_transform` but there will be some other complications. I'll look into this and open up an issue or WIP PR at a later date.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -349,7 +354,7 @@ def get_coefficients(classifier, X_mat):
 ax = sns.distplot(predict_df.query("status == 1").decision_function, hist=False, label='Positives')
 
 
-# In[27]:
+# In[26]:
 
 ax = sns.distplot(predict_df.query("status == 0").probability, hist=False, label='Negatives')
 ax = sns.distplot(predict_df.query("status == 1").probability, hist=False, label='Positives')
```
rdvelazquez: Can we add Y-axis labels, titles and maybe a brief comment or two to help explain these plots? It's not very easy to tell what they are showing, how they are different and how they could be used/interpreted.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -281,66 +300,52 @@ def get_threshold_metrics(y_true, y_pred):
...
-
 predict_df = pd.DataFrame.from_items([
-    ('sample_id', X_all.index),
-    ('testing', X_all.index.isin(X_test_index).astype(int)),
-    ('status', y_all),
-    ('decision_function', final_pipelines[model].decision_function(X_all)),
-    ('probability', final_pipelines[model].predict_proba(X_all)[:, 1])
+    ('sample_id', X.index),
+    ('testing', X.index.isin(X_test.index).astype(int)),
+    ('status', y),
```
rdvelazquez: May want to change the "testing" header to something like "in_testing_set". When I first saw this I thought the testing column was saying if the mutation was predicted as a 0 or 1 by the classifier.

rdvelazquez (commented):
These comments are not specific to this commit but more just recommendations for making the notebook a little clearer.  @patrick-miller feel free to address some/all/none of these (I could open up another PR later to address these) but I think we are using this as a general review of this notebook as it will be used as an important reference/template moving forward.

scripts/2.TCGA-MLexample-covariates.py
rdvelazquez: I didn't realize you need to individually add each comment in a review session (I was assuming they were all tracked/saved for some reason). Lesson learned! My additional comments are below.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: We don't need prepocessing anymore.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: We could include a commented out piece of code for the case where the number of mutations is too low? @dhimmel

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: I agree with this sentiment, and am curious to find out if this is actually important to a biologist.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: These definitely need to be reworked (#97). I'm not a fan of the decision_function plot, but the probability one is more interpretable.

scripts/2.TCGA-MLexample-covariates.py
dhimmel: I'd do something more along the lines of:

```python
pandas.crosstab(coef_df.model, numpy.sign(coef_df.weight))
```

See https://pandas.pydata.org/pandas-docs/stable/generated/pandas.crosstab.html. There are several args you can play with... `normalize`, `margins`

scripts/2.TCGA-MLexample-covariates.py
dhimmel: IIRC the comment may be along the lines of, when n_positives is small, `test_size` may have to be larger so there are a meaningful number of testing positives. Perhaps remove the comment?

If we're not already, printing out an n_positives, n_negatives versus train, test table here would be good.

scripts/2.TCGA-MLexample-covariates.py
dhimmel: > if this is actually important to a biologist.

Yes!

scripts/2.TCGA-MLexample-covariates.py
dhimmel: > I'm not a fan of the decision_function plot, but the probability one is more interpretable.

Yep we had the decision_function plot for back when some classifiers didn't give probabilities. Let's remove it and keep just the probability one

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -79,34 +78,43 @@
...
 
-# Typically, this can only be done where the number of mutations is large enough
+
+# ## Set aside 10% of the data for testing
+
+# In[11]:
+
+# Typically, this type of split can only be done 
+# for genes where the number of mutations is large enough
+X = pd.concat([covariates, expression], axis=1)
```
dhimmel: let's use `axis='columns'` since it's more explicit.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -324,38 +330,44 @@ def positive(w): return '{:,}'.format(np.sum(w > 0))
...
 predict_df['probability_str'] = predict_df['probability'].apply('{:.1%}'.format)
 
 
-# In[24]:
+# In[25]:
 
 # Top predictions amongst negatives (potential hidden responders)
-predict_df.sort_values('decision_function', ascending=False).query("status == 0").head(10)
+model = 'full'
+predict_df.sort_values('decision_function', ascending=False).    query("status == 0 & feature_set == '{0}'".format(model)).head(10)
```
dhimmel: Use newlines / chaining syntax:

```
(predict_df
    .sort_values('decision_function', ascending=False)
    .query("status == 0 and feature_set == @model")
    .head
)
also see how I rewrote query expression
```

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -74,29 +74,26 @@
 
 # In[9]:
 
-# Here are the percentage of tumors with TP53
-y.value_counts(True)
-
-
-# In[10]:
-
 print('Gene expression matrix shape: {0[0]}, {0[1]}'.format(expression.shape))
```
dhimmel: Could we do `*expression.shape` and remove everything inside the brackets?

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -221,7 +218,7 @@ def select_feature_set_columns(X, feature_set):
     cv_results_df = cv_results_df.append(df)
 
 
-# In[18]:
+# In[17]:
 
 # Cross-validated performance heatmap
```
dhimmel: I like the decision to plot model by alpha. I don't think a heatmap is the right visualization since model is categorical. I'd say alpha on x-axis and performance on y-axis, with lines colored by model. (Preferably same color as ROC curve).

This is likely scope creep, so feel free to save this for a future PR.

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -312,16 +309,13 @@ def get_threshold_metrics(y_true, y_pred):
...
 
-coef_df.groupby('feature_set')['weight'].agg([zero, negative, positive])
+print('Signs of the coefficients')
+pd.crosstab(coef_df.feature_set, np.sign(coef_df.weight))
 
 
-# In[23]:
+# In[22]:
 
 model = 'full'
```
dhimmel: I don't see any benefit to defining this model variable. Just use `'full'` inline?

scripts/2.TCGA-MLexample-covariates.py
```diff
@@ -347,27 +341,25 @@ def positive(w): return '{:,}'.format(np.sum(w > 0))
...
+    .query("status == 0 and feature_set == @model")
+    .head(10)
+)
 
 
-# In[26]:
+# In[25]:
 
 # Ignore numpy warning caused by seaborn
 warnings.filterwarnings('ignore', 'using a non-integer number instead of an integer')
```
dhimmel: Check if this is still needed. I think it may be able to be removed due to environment updates.

scripts/2.TCGA-MLexample-covariates.py
patrick-miller: Totally agree. Think it should be separate.  (self-reminder #97)

dhimmel:
@patrick-miller let me know when you're done with edits / think it's mergeable

patrick-miller:
It's good to go. The graphs need to be cleaned up, but that's not related to the pipeline.

dhimmel (approved):

dhimmel:
Thanks @patrick-miller and @rdvelazquez. Really excited for this major advance in the core notebook!

@rdvelazquez now you can work on modifying the notebook to use `dask-searchcv` as per https://github.com/cognoma/machine-learning/issues/94

rdvelazquez:
This was all @patrick-miller. Great job man!

I'll work on dask-searchcv early this week.

PR #103, opened 2017-06-18
Update repository environment, notebooks & READMEs

dhimmel:
Several disjoint changes in a single PR... sorry!

@patrick-miller and @rdvelazquez: do you want to review?


patrick-miller:
Sure, I'll take a look in the next couple of days.

rdvelazquez:
I'll take a look as well.

environment.yml
```diff
@@ -1,17 +1,18 @@
...
 - jupyter=1.0.0
-- numexpr=2.6.0
-- numpy=1.11.1
-- pandas=0.18.1
-- python=3.5.2
-- requests=2.11.1
-- scikit-learn=0.18.0
-- scipy=0.17.1
-- seaborn=0.7.0
+- nomkl
```
rdvelazquez: I get the following error on my 64-bit Windows 7 PC:
```
NoPackagesFoundError: Package missing in current win-64 channels:
  - nomkl
```
I found this [issue](https://github.com/ContinuumIO/anaconda-issues/issues/656) on Continuum's github page.

Changing the line to `conda-forge::nomkl` didn't work.

It may also be of note that when I remove that line, "- nomkl", from `environment.yml` and run `conda env update --file environment .yml` again it works fine and the first line after "Solving package" is "mkl-2017.0.1-0 100%...".

scripts/2.mutation-classifier.py
```diff
@@ -181,31 +188,30 @@ def select_feature_set_columns(X, feature_set):
...
-# In[15]:
+# In[14]:
 
 # Best Parameters
 for model, pipeline in cv_pipelines.items():
-    print('{0}: {1:.3%}'.format(model, pipeline.best_score_))
-
+    print('#', model)
     print(pipeline.best_params_)
+    print('cv_auroc = {:.3%}'.format(pipeline.best_score_))
```
rdvelazquez, line 215: @patrick-miller 's [notebook](https://github.com/cognoma/machine-learning/blob/master/2.TCGA-MLexample-covariates.ipynb) selected `50` as the best parameter for `features__expressions__pca__n_components` for both the full and expression models whereas this notebook is selecting `100` `features__expressions__pca__n_components` for the full model...

Is there a source of randomness that we haven't accounted for by setting the random state? The only thing I can think of is pca which, acording to the [docs](http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html), has a `random_state` property.

rdvelazquez (commented):
Looks good @dhimmel. I read through everything, tried to update my environment , and ran notebook 2. I only have two comments (below). 

I had never seen semantic linefeeds before. Any resources on how that works?

scripts/2.mutation-classifier.py
dhimmel: Great observation. @rdvelazquez it's possible this could happen from package upgrades. But I think the most likely culprit is the randomness in `GridSearchCV` is not controlled. See https://github.com/scikit-learn/scikit-learn/issues/2046.

Possible solutions:

1. set a random seed for numpy
2. construct a  [`StratifiedKFold`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.StratifiedKFold.html#sklearn.model_selection.StratifiedKFold) object with a set `random_state`

I'm leaning towards option 2. I wonder if right now each parameter combination is getting evaluated on different CV splits. I don't see any reason not to reuse CV folds. In fact, it may make for a more consistent comparison.

dhimmel:
> I had never seen semantic linefeeds before. Any resources on how that works?

@rdvelazquez see this [blog post](http://rhodesmill.org/brandon/2012/one-sentence-per-line/). Git tracking and diffs work best with newlines. Accordingly, many people use word wrap at a predetermined max character length per line. Instead, you can put a newline after each sentence. I've found this is a more elegant solution and is more maintainable.

scripts/2.mutation-classifier.py
rdvelazquez: Makes sense to me. I think `StratifiedKFold` will be especially useful for unbalanced genes. I'm in favor of option 2 as well.

Regarding reusing CV folds. I'm not sure if you were talking about using the same folds for each model (expression/covariates/full) or the same folds for each parameter combination within an individual model. If you were referring to the later...  I think this is one of the ways `dasksearchcv` speeds up gridsearch. 

On the topic of `dasksearchcv`: I'll start the `dasksearchcv` PR after this PR is merged.

rdvelazquez:
Thanks for the link! That explained it well.

environment.yml
dhimmel: So the error I'm getting on linux without `nomkl`, i.e. with `mkl` is a dead kernel in the notebook and the error:

```
Intel MKL FATAL ERROR: Cannot load libmkl_avx2.so or libmkl_def.so.
```

Will look into alternative solutions. Also cross-referencing https://github.com/ContinuumIO/anaconda-issues/issues/720, https://github.com/BVLC/caffe/issues/3884, https://github.com/scikit-learn/scikit-learn/issues/5046

dhimmel: @rdvelazquez I think my error was because all my packages, besides a few (particularly `mkl`), were getting installed from the `conda-forge` channel. Specifically, scikit-learn, numpy, and scipy were coming from conda-forge and were the OpenBlas versions rather than the MKL versions. Therefore, I specified the channels for every package in https://github.com/cognoma/machine-learning/pull/103/commits/befb98dedfdc8654c5cb2f4b1e34f6f3bba66cf8.

Now frustratingly, if I `conda env remove --name=cognoma-machine-learning` and then re-install the environment, I keep getting the packages from `conda-forge`. If I rename the `environment.yml` to a different environment name, things work. So somehow my environment is persisting after removal. Frustrating, but I think we can solve later.

It's possibly related the the following messages when I run `conda env remove`:

```
/home/dhimmel/anaconda3/envs/cognoma-machine-learnings/bin/.vega-pre-unlink.sh: line 1: /home/dhimmel/anaconda3/envs/cognoma-machine-learnings/bin/jupyter-nbextension: No such file or directory

WARNING conda.core.link:run_script(519): pre-unlink script failed for package conda-forge::vega-0.4.4-py35_1
consider notifying the package maintainer
```

scripts/2.mutation-classifier.py
patrick-miller: This looks good, it reads more smoothly.

patrick-miller (approved):
The notebook looks good, and it reads much more smoothly now. I'm glad that we switched to a more deliberate creation of the folds, as it will be more consistent with any other methodology we have in the highly unbalanced cases.

PR #109, opened 2017-07-16
Get gene coefficients

rdvelazquez:
Builds on PR #105 

- Adds a function to utils.py, `get_gene_coefficients`, for getting the coefficients for the individual genes from a classifier trained on the principal components returned by pca
- Updates notebook 2 to use this function to output the top 10 genes from the expressions model
- I also added an output for the coefficients from the covariates model for completeness/consistancy so the notebook now shows the top 10 features for each of the three models

@patrick-miller I think your `get_model_coefficients` function from utils is different enough from my `get_gene_coefficients` function that it doesn't make sense to try to combine them... agree? Also, as utils.py is your creation, any input on the style/formatting of my function is very welcome and appreciated.


scripts/2.mutation-classifier.py
```diff
@@ -326,19 +326,45 @@ def get_threshold_metrics(y_true, y_pred):
...
-)
+coef_df.query("feature_set == 'covariates'").head(10)
 
 
-# ## Investigate the predictions
+# ### Top coefficients for full model
 
 # In[22]:
 
+coef_df.query("feature_set == 'full'").head(10)
```
patrick-miller: These updates look good to me. The notebook looks good with how it shows both the covariate loadings, the PCA loadings, and then the weights to the genes.

utils.py
```diff
@@ -102,4 +102,40 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+                           expression_df):
+    """Identify gene coefficients from classifier after pca.
+
+    Args:
+        pca_object: The pca object from running pca on the expression_df.
+        classifier_object: The logistic regression classifier object.
+        expression_df: The original (pre-pca) expression data frame.
+
+    Returns:
+        gene_coefficients: A dataframe with entreze gene-ID, coefficient and
```
patrick-miller: Do we want to also return the name of the gene, instead of just the gene-ID? It would definitely make it more interpretable to the end user.

utils.py
```diff
@@ -102,4 +102,40 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+
+    # Get the classifier coefficients.
+    coefficients = classifier_object.coef_
+    coefficients = pd.DataFrame(coefficients).T
+    # Get the pca weights.
+    weights = pca_object.components_
+    headers = list(expression_df)
+    weights = pd.DataFrame(weights, columns=headers)
+    # Combine the coefficients and weights.
+    combined_df = pd.DataFrame(weights.values*coefficients.values,
```
patrick-miller: You could use numpy.dot here, but this works as well.

patrick-miller (commented):
This looks good to me. I only have one comment about including gene name as well.

utils.py
rdvelazquez: I agree, including the gene name (symbol) would be good. 
@dhimmel any advice on how to do this... I saw that you did something similar in [notebook 3](https://github.com/cognoma/machine-learning/blob/master/3.pathway-classifier.ipynb) (see cell 41), but I'm not sure how.

utils.py
rdvelazquez: Good tip. I'll try np.dot next time.

utils.py
dhimmel: Yes, gene names would be really helpful.

You can get them from https://github.com/cognoma/cancer-data/raw/1d961fb2201209b3ec4774aaadd36f36cf6d165e/data/expression-genes.tsv. You may even want to add the gene description in addition to symbol.

utils.py
dhimmel: @patrick-miller can you explain more how matrix multiplication would be used here?

Also note that python has a matrix multiplication operator since 3.5 (`@`), which perhaps you could use.

utils.py
```diff
@@ -102,4 +102,40 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+                               columns=weights.columns, index=weights.index)
+    gene_coefficients = combined_df.sum(axis=0)
+    gene_coefficients = gene_coefficients.to_frame()
+    gene_coefficients.columns = ['weight']
```
dhimmel: would `gene_coefficients.to_frame(name='weight')` make this line unnecessary?

utils.py
```diff
@@ -102,4 +102,40 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+    # Add absolute value and sort by highest absolute value.
+    gene_coefficients['abs'] = gene_coefficients['weight'].abs()
+    gene_coefficients.sort_values(by='abs', axis=0, ascending=False,
```
dhimmel: `axis=0` is default

utils.py
```diff
@@ -102,4 +102,40 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+    coefficients = classifier_object.coef_
+    coefficients = pd.DataFrame(coefficients).T
+    # Get the pca weights.
+    weights = pca_object.components_
+    headers = list(expression_df)
+    weights = pd.DataFrame(weights, columns=headers)
+    # Combine the coefficients and weights.
+    combined_df = pd.DataFrame(weights.values*coefficients.values,
+                               columns=weights.columns, index=weights.index)
+    gene_coefficients = combined_df.sum(axis=0)
```
dhimmel: I have a preference for `gene_coefficient_df` as the name for the dataframe.

scripts/2.mutation-classifier.py
```diff
@@ -326,19 +326,45 @@ def get_threshold_metrics(y_true, y_pred):
...
+
+
+# ### Top coefficients for individual _genes_ for expressions model
```
dhimmel: Aren't we equally / more interested in the coefficients from the `full` model?

I think it would be really useful to show the gene coefficients from both models with expression in the same table. You could sort by abs coefficient of the full model, but the comparison between coefficients will be nice.

scripts/2.mutation-classifier.py
dhimmel: Are the PCA components the same between the `expression` and `full` models? I wonder if showing in one dataframe makes sense for comparison. I'm looking to get at the question, what effect does including covariates have on coefficients.

dhimmel (commented):
Nice PR. Your coding style looks to be improving quickly!

utils.py
rdvelazquez: Makes sense. That's the only dataframe that isn't named with a `..._df`. Good catch!

scripts/2.mutation-classifier.py
rdvelazquez: Let me know if my comment on the number of PCA components used by each model changes your preferred approach for this. I'm thinking creating one combined dataframe with coefficients and absolute values for both `full` and `expression` and also including two columns for "rank" for full and expression and then showing the top 10 genes for each model as two separate outputs may be the best bet. That may take me a while.

I'll look into it a little more tonight or tomorrow but let me know your thoughts.

utils.py
rdvelazquez: Yep. That worked.

scripts/2.mutation-classifier.py
rdvelazquez: The PCA components are the same; however, in this notebook the `full` model selected 50 components as the best parameter for `n_components` and the `expression` model selected 100 components as the best parameter for `n_components`. 

If we were to try and compare gene coefficients between the two models it would be difficult to determine what effect including covariates had on coefficients because the addition of coefficients isn't the only thing that changed... the number of PCA components also changed.

That said, it would likely still be interesting to see which genes were most impactful to the classifier when including covariates (the `full` model). We may want to present the gene coefficients for the full model in a separate table because doing a side by side comparison with the gene coefficients from the expression only model may make people draw conclusions that aren't there.

utils.py
patrick-miller: I might be mistaken, but doesn't 
```
np.dot(weights.values.T, coefficients.values)
```
save a line or two? Or getting rid of some of the pandas in the code:

```
coefficients = classifier_object.coef_
weights = pca_object.components_
gene_coefficients = pd.DataFrame(np.dot(weights.T, coefficients).T)
```

scripts/2.mutation-classifier.py
patrick-miller: I agree that we would be more interested in the coefficients for the `full` model. I think that the `expression` model really just serves as a benchmark for comparison of AUROC.

utils.py
rdvelazquez: Makes sense to me although my linear algebra is a little rusty and I'm not sure I'm actually taking the dot product (I think I may have referred to what I was doing as "dot product" but I'm not sure that was correct).

I'll give this a try tomorrow and let you know if it works.

scripts/2.mutation-classifier.py
dhimmel: > If we were to try and compare gene coefficients between the two models it would be difficult to determine what effect including covariates had on coefficients because the addition of coefficients isn't the only thing that changed

Good point. Given this consideration my opinion is:

1. In the case of gene-level coefficients, a combined dataframe is most useful.
2. In the case of component-level coefficients, a combined dataframe is less essential but still likely preferable. Of course, you can note that different models may have received different numbers of components. And some coefficient values will be missing for some models (since those components were not evaluated by the logistic regression).

However, ease of implementation is also important. I don't want to burden the notebook with lot's of code to get these consolidated coefficient dfs. But I do think they will be most useful.

utils.py
rdvelazquez: It worked!

I replaced:
```
combined_df = pd.DataFrame(weights.values*coefficients.values,
                           columns=weights.columns, index=weights.index)
gene_coefficients = combined_df.sum(axis=0)
gene_coefficients = gene_coefficients.to_frame()
```
with:
```
    gene_coefficients = np.dot(weights.values.T, coefficients.values)
    gene_coefficients_df = pd.DataFrame(gene_coefficients, columns=['Weight'], index=weights.columns)
```
It was a little long and hard to follow as one line so I split it into two.

dhimmel: Does `weights.values.T @ coefficients.values` work?

patrick-miller: I think it is better style to skip the pandas DataFrame step altogether.

`coefficients = pd.DataFrame(coefficients).T`
and
`weights = pd.DataFrame(weights, columns=headers)`

are overkill, in my opinion, since you are just extracting the `.values` anyway.

rdvelazquez:
@dhimmel and @patrick-miller my last commit addresses most of your comments. I just haven't combined the coefficient dataframes for `full` and `combined`. 

I thought we had removed all the sources of randomness in this notebook to make it reproducible but it this time when I ran the notebook it only selected `50` for `n_components` for both the `full` and the `expression` models whereas previous runs of the notebook selected `100` for the `expressions` model.

rdvelazquez:
FYI - I just saw the recent comments on the dot product. The recent commit doesn't address either of those.

rdvelazquez:
One more update commit coming soon (hopefully). This isn't quite ready for review yet.

rdvelazquez:
This is now ready to review.

I was able to get rid of the unnecessary pandas dataframes as @patrick-miller suggested and I used the new `@` matrix multiplication (new to python 3.5) as @dhimmel  suggested. Thank you both for the detailed review!! That type of input is immensely helpful for improving my coding. 

I didn't combine the coefficients dataframes for the `full` and `expression` models. @dhimmel let me know if this is something you think we should do. There is already quite a lot in the notebook regarding coefficients so I see where you are coming from with not wanting to bog the notebook down with too much code.

utils.py
```diff
@@ -102,4 +102,54 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+    # Add gene symbol and description
+    gene_coefficients_df['Symbol'] = expression_genes_df['symbol']
+    gene_coefficients_df['Description'] = expression_genes_df['description'] 
+    # Add absolute value and sort by highest absolute value.
+    gene_coefficients_df['Abs'] = gene_coefficients_df['Weight'].abs()
+    gene_coefficients_df.sort_values(by='Abs', ascending=False, inplace=True)
+    # Reorder columns
+    gene_coefficients_df = gene_coefficients_df[['Symbol', 'Weight', 'Abs',
+                                                'Description']]
+    return(gene_coefficients_df)
```
dhimmel: Add blank newline at end of file.

scripts/2.mutation-classifier.py
```diff
@@ -320,24 +325,75 @@ def get_threshold_metrics(y_true, y_pred):
...
+    final_pipelines['full']
+    .named_steps['classify']
+    )
+num_covariates = len(covariate_df.columns)
+gene_coefficients_for_full = get_genes_coefficients(
+    pca_object=pca_for_full,
+    classifier_object=classifier_for_full,
+    expression_df=expression_df,
+    expression_genes_df=expression_genes_df,
+    num_covariates=num_covariates
```
dhimmel: Let's just put `len(covariate_df.columns)` here rather than assigning it to `num_covariates`

utils.py
```diff
@@ -102,4 +102,54 @@ def get_model_coefficients(classifier, feature_set, covariate_names):
...
+    weights = pca_object.components_
+    # Combine the coefficients and weights.
+    gene_coefficients = weights.T @ coefficients.T
+    # Create the dataframe with correct index
+    gene_coefficients_df = pd.DataFrame(gene_coefficients, columns=['Weight'])
+    gene_coefficients_df.index = expression_df.columns
+    gene_coefficients_df.index.name = 'Entrez ID'
+    expression_genes_df.index = expression_genes_df.index.map(str)
+    # Add gene symbol and description
+    gene_coefficients_df['Symbol'] = expression_genes_df['symbol']
```
dhimmel: Let's stick with the `lowercase_underscore_separated` column name convention.

dhimmel (commented):
@rdvelazquez great work! Almost ready to merge. Just a few minor comments

rdvelazquez:
@dhimmel my last commit should address your three comments. I added a blank new line to the end of the utils.py file but I don't see it when I view the file on github... I see it when I view it in jupyter and IDLE.

dhimmel:
>  I added a blank new line to the end of the utils.py file but I don't see it when I view the file on github

That's right but the missing newline symbol on GitHub that used to be there is gone.

dhimmel (approved):

PR #111, opened 2017-08-21
Allow for environment variables in notebook

patrick-miller:
This pull request seeks to merge the workflow in [ml_workers](https://github.com/cognoma/ml-workers/blob/master/notebooks/2.mutation-classifier.ipynb) so that we can more easily move our notebook from this repo to production.

Let me know what you think.


scripts/2.mutation-classifier.py
```diff
@@ -36,8 +38,28 @@
...
+# We're going to be building a classifier with multiple genes filtered by two diseases 
+# Example:
+# gene_ids = ['7157', '7158', '7159', '7161']
+
+# Information passed into the notebook is stored in environment variables
+gene_ids = os.environ.get('gene_ids')
+if not gene_ids:
+    gene_ids = ['7157'] # TP53 is the default
+else:
+    gene_ids = gene_ids.split('-')
```
dhimmel, line 51: I think comma separating makes more sense than hyphen. @dcgoss how hard would it to switch to commas?

scripts/2.mutation-classifier.py
```diff
@@ -36,8 +38,28 @@
...
+    
+disease_acronyms = os.environ.get('disease_acronyms')
+
+if not disease_acronyms:
+    disease_acronyms = [] # use all of the diseases as default
```
dhimmel: I think it makes sense to keep `disease_acronyms = None` for the case where all diseases are used...

scripts/2.mutation-classifier.py
```diff
@@ -48,46 +70,76 @@
...
 # In[5]:
 
-get_ipython().run_cell_magic('time', '', "path = os.path.join('download', 'mutation-matrix.tsv.bz2')\nmutation_df = pd.read_table(path, index_col=0)")
+
+# Select acronym_x and n_mutations_log1p covariates only
+disease_cols = [col for col in covariate_df.columns if col.startswith('acronym_')]
+
+# Filter covariate columns by disease if a list was provided
+if disease_acronyms:
+    disease_cols = [col for col in disease_cols if col.endswith(tuple(disease_acronyms))]
```
dhimmel, line 95: Woah! Didn't know that in [`str.endswith`](https://docs.python.org/3/library/stdtypes.html), "suffix can also be a tuple of suffixes to look for". That is super convenient!

scripts/2.mutation-classifier.py
```diff
@@ -48,46 +70,76 @@
...
+    
+selected_cols = disease_cols + ['n_mutations_log1p']
+covariate_df = covariate_df[selected_cols]
 
 
 # In[6]:
 
+
 get_ipython().run_cell_magic('time', '', "path = os.path.join('download', 'expression-genes.tsv')\nexpression_genes_df = pd.read_table(path, index_col=0)")
```
dhimmel: Let's also get rid of `%%time` here. And consider moving this to the load data cell.

dhimmel (commented):
Really nice. Some small notes, but looks great.

dhimmel (approved):

dhimmel:
@patrick-miller I gave you write access to this `cogonoma/machine-learning`. Want to "Squash and merge" this PR?

Issue #110, opened 2017-08-16
Machine Learning Punch List for Launch


rdvelazquez:
Here's the general punch list that we discussed at tonight's meetup for getting the machine learning part of cognoma launch ready.

- [x] Update Notebook 2 to be compatible with ml-workers and PR ml-workers so the two notebooks are the same (https://github.com/cognoma/machine-learning/pull/111 by @patrick-miller)
- [x] Change plotting in notebooks from vega to plotnine (https://github.com/cognoma/machine-learning/pull/112 by @patrick-miller)
- [x] Look into the [memory usage issues](https://github.com/cognoma/core-service/issues/88) (@wisygig)
- [x] Look into the [number of pca components to use](https://github.com/cognoma/machine-learning/issues/106) (#113 and #114 by @rdvelazquez)
- [x] Optimize between memory usage, pca components (and the regulirization_alpha_list), and how large of an AWS instance we use (@dhimmel)

To be completed at a later date: Templating for jupyter [notebooks](https://github.com/cognoma/ml-workers/pull/13) (@wisygig)

patrick-miller:
Once we are happy with the final notebook, we should do a final cleanup (there is still a warning in the import section) and then I will do a PR into `mlworkers`.

patrick-miller:
@rdvelazquez @dhimmel @wisygig

Because the launch party is coming up, I was going to open a PR to push the notebook to production this week/weekend. I think Ryan has made very good progress, and our focus should probably be put towards making sure that his changes work on our AWS instances.

dhimmel:
> Because the launch party is coming up, I was going to open a PR to push the notebook to production this week/weekend

I agree & fully support!

rdvelazquez:
> I was going to open a PR to push the notebook to production this week/weekend.

Sounds good to me!

patrick-miller:
Using the updated notebook and a small sample set, I got reasonable results back in production. We are getting the following warning for a pandas import:

![image](https://user-images.githubusercontent.com/5490674/31202431-9bd55b70-a930-11e7-8801-b5b485d29a29.png)

I'm going to run a larger query, to see if we get memory issues.

patrick-miller:
The larger query for TP53 as the gene and all diseases selected runs into a memory error. 

@dhimmel @rdvelazquez

dhimmel:
> The larger query for TP53 as the gene and all diseases selected runs into a memory error.

[...]

PR #112, opened 2017-08-22
Change plots to use plotnine

patrick-miller:
Here I am replacing all of the plots with the ggplot inspired `plotnine` package. I think that these look good to me, but let me know what you think.


rdvelazquez:
Nice work. These look really good based on my first look at them. A few quick comments:

1. Should you also revise the utils.py file to remove the functions that we no longer need?

2. I really like the revision to the probabilities plots at the end of the notebook (showing the plots for all three models). I think labeling the axis would also be very helpful.

scripts/2.mutation-classifier.py
```diff
@@ -17,27 +18,45 @@
 from dask_searchcv import GridSearchCV
 from sklearn.pipeline import Pipeline, FeatureUnion
 from sklearn.preprocessing import StandardScaler, FunctionTransformer
-from vega import Vega
-import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
-import seaborn as sns
+from plotnine import *
```
dhimmel: Let's avoid `import *` ([reasons](https://stackoverflow.com/a/2386740/4651668)). If you're worried about plotnine taking up lot's of space you could do something like `import plotnine as pln` or `import plotnine as gg`.

scripts/2.mutation-classifier.py
```diff
@@ -234,21 +289,26 @@ def select_feature_set_columns(X, feature_set):
...
-                              columns='classify__alpha')
-ax = sns.heatmap(cv_score_mat, annot=True, fmt='.1%')
-ax.set_xlabel('Regularization strength multiplier (alpha)')
-ax.set_ylabel('Feature Set');
+cv_results_df['classify__alpha'] = cv_results_df['classify__alpha'].astype(str)
+
+(ggplot(cv_results_df, aes(x='classify__alpha',
+                           y='mean_test_score',
+                           fill='feature_set'))
+ + geom_bar(stat='identity', position='dodge')
```
dhimmel: I'm not sure a barplot is the best. I'm thinking a scatterplot with model shown via line/dot color or via facet.

You could do a [`scale_x_log10`](http://plotnine.readthedocs.io/en/stable/generated/plotnine.scales.scale_x_log10.html) so the x-axis is continuous rather than discreet.

Ideally plot would expand_limits for y to include 0.5 and 1. This will have to wait until the next plotnine release: see https://github.com/has2k1/plotnine/commit/192609d761726c84ffec38dcbade325d9556364e. In other words, 0.5-1.0 should be the ylim unless AUROCs are below 0.5 in which case the lower bound should accommodate the lowest AUROC.

scripts/2.mutation-classifier.py
```diff
@@ -295,20 +356,31 @@ def get_threshold_metrics(y_true, y_pred):
...
+
+# ### AUROC
 
-final_spec = fill_spec_with_data(vega_spec, 
-    {'roc': roc_output, 'legend_auc': auc_output})
+# In[21]:
 
-Vega(final_spec)
+
+pd.pivot_table(auc_output,
```
dhimmel: Minor: models as rows and partition as columns makes a little more sense to me.

dhimmel (commented):
Nice. What do you think about the default plotnine style? In general, I find `theme_bw` to be more aesthetic. If we want, we can define a style in utils. Always can do this in a future PR. Here are some [ggplot2 mods](https://github.com/dhimmel/hetior/blob/e24db93e99fb75085d6b186ecf052cba516c05e0/R/plotting.R) I often use.

scripts/2.mutation-classifier.py
patrick-miller: Agreed, I was just copying from some `plotnine` examples.

scripts/2.mutation-classifier.py
patrick-miller: I switched to a scatterplot and the log scale -- it initially looked funky to me with the default theme. I used a hack to get 'expand limits'.

scripts/2.mutation-classifier.py
patrick-miller: Done

patrick-miller:
I translated one of my R ggplot themes to `plotnine` in the utils.py file. We can play around with this and the color scheme for the plots. The ROC plot might be a little busy.

rdvelazquez:
Would it be possible to show the actual AUROC values on the AUROC plot? Maybe next to the lines on the legend? If not maybe we can insert a table that shows this. I think this is one of the more important pieces of information in the notebook so it would be good to be able to easily see the testing and training AUROC for each model.

dhimmel (approved):

dhimmel:
@rdvelazquez I gave you write access to this repo. After all your suggestions have been addressed and @patrick-miller is done, you can "Squash and merge" this pull request. Make sure to select "Squash and merge" rather than "Create a merge commit"

patrick-miller:
I agree @rdvelazquez about the AUROC values. I include them in the next few cells, but it would be better to have them inside the plot itself. Because there are 6 values, things can get cluttered. Maybe we only include the test AUROC inside the plot?

dhimmel:
I'm fine with the current display of AUROCs (in the following cell). I don't think there is really a clean way to include them in the plot.

utils.py
```diff
@@ -1,69 +1,22 @@
...
+          line = gg.element_line(color = "#4d4d4d"), 
+          rect = gg.element_rect(fill = "white", color = None), 
+          text = gg.element_text(color = "black"), 
+          axis_ticks = gg.element_line(color = "#4d4d4d"),
+          legend_key = gg.element_rect(color = None), 
+          panel_border = gg.element_rect(color = "#4d4d4d"),  
+          panel_grid = gg.element_line(color = "#b3b3b3"), 
+          panel_grid_major_x = gg.element_blank(),
+          panel_grid_minor = gg.element_blank(),
+          strip_background = gg.element_rect(fill = "#e5e5e5", color = None),
```
dhimmel: Consider changing `#e5e5e5` to `#FEF2E2` for `strip.background` color to make the facet strips a little less drab.

patrick-miller: Will do

dhimmel:
@patrick-miller merge when you're ready. So ml-workers should be able now to directly use this notebook?

@dcgoss how should we go about using this notebook in production? Can ml-workers just use the URL for this notebook once it's merged?

rdvelazquez:
> how should we go about using this notebook in production? Can ml-workers just use the URL for this notebook once it's merged?

We are planning to use the machine-learning repo for development of the notebooks but ml-workers for production so the next step (either now or once all the items in #110 are done) will be to open a PR in ml-workers updating that notebook. @patrick-miller 's revisions in #111 were designed to let this notebook work in both the machine-learning and the ml-workers repos. 

I'm not opposed to just having ml-workers reference this notebook but I think when we discussed it at the last meetup there were some potential issues with doing that.

